PR #2, opened 2017-05-30
[IPAAS-287] Tests to verify advanced integration with multiple steps

lvydra:


e2e/app.po.ts
```diff
@@ -103,7 +103,9 @@ export class AppPage {
 
   clickButton(buttonTitle: string): P<any> {
     log.info(`clicking button ${buttonTitle}`);
-    return this.getButton(buttonTitle).click();
+    const buttonElement = this.getButton(buttonTitle);
+    this.waitForElementToBePresent(buttonElement);
+    return buttonElement.click();
   }
```
dsimansk: +1

e2e/app.po.ts
```diff
@@ -195,6 +197,16 @@ export class AppPage {
...
+  waitForElementToBePresent(element) {
+    browser.wait(function () {
+        return element.isPresent();
+    }, 60000);
+
+    browser.wait(function () {
+        return element.isDisplayed();
+    }, 60000);
+  }
+
```
dsimansk: I like the idea to wait for element in generatl. Maybe we could wrap this with `ExpectedConditions` and use either [`elementToBeClicable()`](https://github.com/angular/protractor/blob/master/lib/expectedConditions.ts#L188) or [`visibilityOf`](https://github.com/angular/protractor/blob/master/lib/expectedConditions.ts#L393). Wdyt?
Cc @jludvice

lvydra: Sorry, I didn't notice we have functionality for that. 
What about that https://github.com/lvydra/syndesis-e2e-tests/commit/594ea73f0f897cb5dd0b56bf262e7990b3b8a9d1

e2e/app.po.ts
```diff
@@ -103,7 +103,9 @@ export class AppPage {
 
   clickButton(buttonTitle: string): P<any> {
     log.info(`clicking button ${buttonTitle}`);
-    return this.getButton(buttonTitle).click();
+    const buttonElement = this.getButton(buttonTitle);
+    browser.wait(ExpectedConditions.presenceOf(buttonElement), 10000, 'Wait for button');
```
dsimansk: Lets change this to `visibilityOf()` as it's a bit stronger. As it represents: "element is present in DOM && visible". Sorry for nit-picking here :)

jludvice: I'm wondering what happens if `browser.wait` times out (in the context of https://github.com/syndesisio/syndesis-ui/issues/559).
We probably should return rejected promise with error data.
@lvydra can you check what happens if you wait for non-existent button? Is cucumber report generated or does it crash protractor and leave browser open ?

lvydra: OK, I will check it.

lvydra: I test it and it just crash and leave browser open.

dsimansk:
@lvydra could you please try to implement the button wait in similar fashion. Seems to be working for me. Tests are finished with failure and report is generated. 
```
clickButton(buttonTitle: string): P<any> {
    log.info(`clicking button ${buttonTitle}`);
    return browser.wait(ExpectedConditions.visibilityOf(this.getButton(buttonTitle)), 6000, 'No button visible')
      .then(() =>  this.getButton(buttonTitle).click())
      .catch((e) => P.reject(e) );
  }
```

lvydra:
OK

lvydra:
Done plus "random" choosing from multiple links with same title added.

PR #3, opened 2017-05-31
ci: remove pipeline library declaration

dsimansk:
This declaration is not need anymore as we have pipeline declared in job itself.


dsimansk:
We need the library, just need to get rid off direct declaration in Jenkinsfile, because of changed job configuration itself. Library is still provided by in a different way now.

PR #5, opened 2017-06-05
fix: use non-dsl arq variant in pipeline

dsimansk:
WIP, do not merge, if the PR job fails.


dsimansk:
@jludvice self-merging this one now!

PR #18, opened 2017-07-17
fix: Pass variables to container

dsimansk:


dsimansk:
Not needed anymore since Jenkins plugins were fixed.

PR #23, opened 2017-07-20
fix(atlasmap): Ignore advanced scenario until root cause is fixed

dsimansk:
Lets ignore currently undoable scenario for Atlasmap UI. Blocked by https://github.com/atlasmap/atlasmap/issues/28


lvydra:
OK

PR #30, opened 2017-07-24
fix: Lock Chrome webdriver version to 2.30

### LINKED ISSUE #29: Lock version on chromedriver to 2.30
It seems that latest version of Chromedriver 2.31 is failing to start browser correctly. We should lock down the version to 2.30 and investigate further.

dsimansk:
Fixes #29


dsimansk:
Pls a quick review, any of you comes first @jludvice @lvydra. I've noticed fails during Chrome start inside Docker container, not reproducible without Docker though.

Issue #43, opened 2017-07-26
Investigate multiple step integration test failure


dsimansk:
As discussed during today's meeting, there's a failure in multiple-step-integration. Pls take a look as a priority.

This also takes down protractor so definitely need a bit of care.

https://jenkins-syndesis-ci.b6ff.rh-idev.openshiftapps.com/job/syndesis-e2e-tests/job/master/32/console

```
[07:02:39] E/launcher - unknown error: Element <span _ngcontent-c7="" class="parent-step ">...</span> is not clickable at point (174, 76). Other element would receive the click: <a _ngcontent-c0="" href="/dashboard">...</a>
  (Session info: chrome=59.0.3071.115)
  (Driver info: chromedriver=2.30.477691 (6ee44a7247c639c0703f291d320bdf05c1531b57),platform=Linux 3.10.0-514.25.2.el7.x86_64 x86_64)
[07:02:39] E/launcher - WebDriverError: unknown error: Element <span _ngcontent-c7="" class="parent-step ">...</span> is not clickable at point (174, 76). Other element would receive the click: <a _ngcontent-c0="" href="/dashboard">...</a>
  (Session info: chrome=59.0.3071.115)
  (Driver info: chromedriver=2.30.477691 (6ee44a7247c639c0703f291d320bdf05c1531b57),platform=Linux 3.10.0-514.25.2.el7.x86_64 x86_64)
    at Object.checkLegacyResponse (/home/jenkins/workspace/syndesis-e2e-tests_master-T3LXHLH4EPAVX4E4OGU73HSCISH2FS2SDVT6KMVDFTLN5LHCVFNA/node_modules/selenium-webdriver/lib/error.js:505:15)
    at parseHttpResponse (/home/jenkins/workspace/syndesis-e2e-tests_master-T3LXHLH4EPAVX4E4OGU73HSCISH2FS2SDVT6KMVDFTLN5LHCVFNA/node_modules/selenium-webdriver/lib/http.js:509:13)
    at doSend.then.response (/home/jenkins/workspace/syndesis-e2e-tests_master-T3LXHLH4EPAVX4E4OGU73HSCISH2FS2SDVT6KMVDFTLN5LHCVFNA/node_modules/selenium-webdriver/lib/http.js:440:13)
    at <anonymous>
    at process._tickCallback (internal/process/next_tick.js:188:7)
From: Task: WebElement.click()
[...]

PR #44, opened 2017-07-26
fix: Hiding navigation bar

lvydra:
#43

### DIFF
--- a/e2e/integrations/integration.steps.ts
+++ b/e2e/integrations/integration.steps.ts
@@ -207,6 +207,14 @@ class IntegrationSteps {
           });
         });
       }
+
+      navBar.isDisplayed().then((visible) => {
+        if (!visible) {
+          const navBarToggle = this.world.app.getElementByClassName('navbar-toggle');
+          navBarToggle.click();
+        }
+      });
+
       return stepsArray;
     });
   }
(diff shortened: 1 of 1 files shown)


e2e/integrations/integration.steps.ts
```diff
@@ -183,6 +183,15 @@ class IntegrationSteps {
...
     const steps = this.world.app.getElementsByClassName('parent-step');
+    const navBar = this.world.app.getElementByClassName('nav-pf-vertical');
+
+    navBar.isDisplayed().then((visible) => {
+      if (visible) {
+        log.warn('Navigation bar is not hidden.');
+        const navBarToggle = this.world.app.getElementByClassName('navbar-toggle');
+        navBarToggle.click();
+      }
+    });
```
dsimansk: I guess now we're permanently hiding the nav bar due to removal of this line https://github.com/syndesisio/syndesis-e2e-tests/pull/44/commits/4de6a0d584c7c3c460a385a6614dcc5156f5c479#diff-b5c6e4c77b8cafd1a1ce1d7a0bc7b3e6R204. Either add back that click with condition also or we need to sanitize navigation bar navigation to check if visible before clicking.

lvydra: Sorry I forgot to add it :-) However I think we should consider adding check to navigation bar navigation.

dsimansk: Np, I've approved this one. Lets call this temp workaround, feel free to create new PR with enhanced navigation bar code :).

PR #47, opened 2017-07-31
ci: Wait for the test namespace to be empty

dsimansk:
@jludvice wdyt too naive and simple?

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,6 +1,14 @@
 def namespace = System.getenv().get('E2E_NAMESPACE')
 
 node {
+  //wait for the test namespace to not exist to prevent conflicts
+  timeout(600) {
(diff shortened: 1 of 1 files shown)


dsimansk:
@jludvice can you recheck it please, my first try was too naive. Now it should return `1` if project doesn't exist and `0` for existing one.

jludvice:
yup, just checked, that

```bash
$ oc get project syndesis-qasdf
No resources found.
Error from server: User "jludvice" cannot get project "syndesis-qasdf"

$ echo $?                      
1

 $ oc get project syndesis-qe   
NAME          DISPLAY NAME   STATUS
syndesis-qe                  Active

$ echo $?                   
0
```

I'm +1 to merge and see how it works

PR #52, opened 2017-08-04
[fix] Fix implementation of scrolling step

jludvice:


dsimansk:
I'll merge this manually as this is simple fix and the tests are currently broken due to #53.

Issue #13, opened 2017-07-17
Implement E2E test for Basic Filter step


dsimansk:
Based on this [Filtering epic issue](https://github.com/syndesisio/syndesis-project/issues/2) create E2E test that will verify addition of filter to integrations. 

Should be doable after PR https://github.com/syndesisio/syndesis-ui/pull/582 is merged.

dsimansk:
I've changed the scope of this task to cover only Basic Filter. There are 2 types of filters now. Basic Filter which was delivered from Spring 14. And Advance Filter that allows to use usual Camel filter expressions, this is planned to be enhanced further with code editor, documentation link and example usage. 

Please prepare only test for Basic Filter as the UI is rather complex, with input boxes etc.

lvydra:
"Add another rule" link is not working on Basic Filter page. Should it be functional?

dsimansk:
I've found the issue reported, but still not resolved https://github.com/syndesisio/syndesis-ui/issues/631. So it should be functional, but it isn't right now :)

lvydra:
After leaving  basic filter step definition and returning back, all fields are reset to initial (blank) state. Is it intentional?

dsimansk:
Done via #62.

PR #62, opened 2017-08-11
[relates #13] Basic filter test

lvydra:
#13

### DIFF
--- a/e2e/integrations/edit/edit.po.ts
+++ b/e2e/integrations/edit/edit.po.ts
@@ -136,8 +136,8 @@ export class StepFactory {
     }
     if (stepType.toUpperCase() === 'LOG') {
       return new IntegrationConfigureLogStepPage(parameter);
-    } else if (stepType.toUpperCase() === 'FILTER') {
-      return new IntegrationConfigureFilterStepPage(parameter);
+    } else if (stepType.toUpperCase() === 'BASIC FILTER') {
+      return new IntegrationConfigureBasicFilterStepPage(parameter);
     }
 
     return null;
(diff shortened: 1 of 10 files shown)


e2e/integrations/integration.steps.ts
```diff
@@ -100,107 +101,118 @@ class IntegrationSteps {
...
 
-    this.getStepsArray().then((array) => {
-      this.world.app.clickButton('Add a Step');
+    const array = await this.getStepsArray();
+
+    this.world.app.clickButton('Add a Step');
+
+    const links = this.world.app.getLinks('Add a step');
+    const count = await links.count();
+
```
dsimansk: Are there links and buttosn that has different capitalization of `Step`? If that's the case, please create new issue in UI repo.

e2e/integrations/edit/edit.po.ts
```diff
@@ -209,52 +209,249 @@ export class IntegrationConfigureLogStepPage extends IntegrationConfigureStepPag
...
+  }
+
+  setOp(op: number): P<any> {
+    log.info(`setting basic filter step predicate to option number ${op}`);
+    const opOptions = this.rootElement().all(by.css(IntegrationConfigureBasicFilterStepPage.opOptionSelector));
+    return opOptions.then((options) => {
+      options[op].click();
+    });
+  }
+
```
dsimansk: `op` is index? I'd probably prefer operation name or some abbreviation rather than index. Purely due to improved readability of code :)

e2e/integrations/integration.steps.ts
lvydra: Yes, there is a button with Step and link with step.

e2e/integrations/edit/edit.po.ts
lvydra: `op` is name of select.

e2e/integrations/integration.steps.ts
dsimansk: Though so, then please open new issue to get it uniformed.

e2e/integrations/edit/edit.po.ts
dsimansk: Works for me now, we can always revisit it later.

PR #74, opened 2017-09-01
[closes #34] Implement basic test for oauth settings page

### LINKED ISSUE #34: Implement test for Global Settings
The preview of the page is available in latest builds and also present on `syndesis-qe` instance. It's a good starting point to begin with e2e test impl.

jludvice:
@mmelko @lvydra @dsimansk 

This is very basic test for settings page.
I didn't do deleting of settings as it throws HTTP500 (reported as syndesis-rest/issues#549)
and it wasn't clear how it was supposed to work.
Delete settings is TODO for future

### DIFF
--- a/e2e/settings/settings.po.ts
+++ b/e2e/settings/settings.po.ts
@@ -1,0 +1,97 @@
+import {SyndesisComponent} from '../common/common';
+import {$, by, ElementArrayFinder, ElementFinder} from 'protractor';
+import {P} from '../common/world';
+import {log} from '../../src/app/logging';
+
+export class SettingsPage implements SyndesisComponent {
+  rootElement(): ElementFinder {
+    return $('syndesis-settings-root');
+  }
+
+  /**
+   * Get title of active settings tab
+   * @returns {Promise<string>} title of active settings tab
+   */
+  async activeTab(): P<string> {
+    const activeTab = this.rootElement().$('ul.nav-tabs li.active a');
+    return activeTab.getText();
+  }
+
+  /**
+   * Support method for finding settings items by name
+   * @param {ElementFinder} parent search child elements of this element
+   * @returns {ElementArrayFinder} found elements
+   */
+  listSettingsItems(parent: ElementFinder = this.rootElement()): ElementArrayFinder {
+    return parent
+      .$('pfng-list')
+      .$$('div.list-pf-item');
+  }
+
+  /**
+   * Fetch all settings items and find proper one according to given name.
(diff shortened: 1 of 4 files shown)


e2e/settings/settings.steps.ts
```diff
@@ -0,0 +1,49 @@
...
+  @when(/^"([^"]*)" clicks to the "([^"]*)" item "([^"]*)" button$/)
+  public clickSettingsButton(userAlias: string, itemTitle: string, buttonTitle: string): P<any> {
+    const page = new SettingsPage();
+    return page.clickButton(itemTitle, buttonTitle);
+  }
+
+  @when(/^fill form in "([^"]*)" settings item$/)
+  public fillSettingsItemForm(itemTitle: string): P<any> {
+    const settings = new OAuthSettingsComponent();
+    const toFill = this.world.testConfig.settings[itemTitle];
```
jludvice: This test requires update to test settings config map.
I already did that on syndesis-ci.

e2e/settings/settings.po.ts
```diff
@@ -0,0 +1,97 @@
...
+    return item.element(by.buttonText(buttonTitle)).click();
+  }
+}
+
+
+/**
+ * Represents oauth apps tab.
+ */
+export class OAuthSettingsComponent extends SettingsPage {
+  rootElement(): ElementFinder {
```
jludvice: I'm not sure inheritance is good approach here.
What do you think about it?

e2e/settings/settings.po.ts
```diff
@@ -0,0 +1,97 @@
...
+  }
+
+  /**
+   * Get title of active settings tab
+   * @returns {Promise<string>} title of active settings tab
+   */
+  async activeTab(): P<string> {
+    const activeTab = this.rootElement().$('ul.nav-tabs li.active a');
+    return activeTab.getText();
+  }
```
jludvice: I hope that other tabs will have similar structure and methods `getSettingsItem` and `clickButton` will work also for them.

e2e/settings/settings.po.ts
lvydra: I'm not sure to. Oauth apps tab as part of setting page, defined by 'syndesis-settings-root' element, could be better as SettingPage attribute.  But still we could use some kind of general ancestor for tabs.

jludvice: @lvydra can we leave it this way for some time?
I bet we'll have to rewrite it anyway once engineering implements another Tab on this page.

PR #89, opened 2017-09-11
fix: Rm navigation to Home via top bar logo

dsimansk:
Caused by https://github.com/syndesisio/syndesis-ui/issues/924


dsimansk:
Not needed anymore, I'm closing it without merging.

PR #91, opened 2017-09-12
fix: Fix missing DOM elements

dsimansk:


dsimansk:
@jludvice I've probably fixed all outstanding issue, pls review it whenever you're available :)

e2e/integrations/integration.steps.ts
```diff
@@ -59,6 +59,10 @@ class IntegrationSteps {
   @when(/^she selects "([^"]*)" integration action$/)
   public selectIntegrationAction(action: string): P<any> {
     const page = new ListActionsComponent();
+    if (action === 'Create Opportunity') {
+      log.warn(`Action ${action} is not available`);
+      return page.selectAction('Create Salesforce object');
```
jludvice: what is this hack? :D

dsimansk: It's kind of long to explain. There're a few occurrences of `Create Opportunity` action in scenarios, but it was removed from the selectable actions. I wanted to keep the scenarios and make the tests work with current state. But there was no value in that action anyway.

dsimansk: @jludvice any more comments. I promise I'll fix it properly later :)

PR #100, opened 2017-10-06
[fix] Tests fix

lvydra:


dsimansk:
There's a failure in Dashboard test that kills Protractor as usual.
```
2017-10-10 08:46:21,703 INFO [root] searching for title Recent Updates
[08:46:21] E/launcher - There must be present a table Top 5 Integrations: expected false to be true
[08:46:21] E/launcher - AssertionError: There must be present a table Top 5 Integrations: expected false to be true
```

e2e/dashboard/dashboard.feature
```diff
@@ -70,9 +70,11 @@ Feature: Test to verify Dashboard links
     And she selects "Mention" integration action
     Then she is prompted to select a "Finish" connection from a list of available connections
 
-    When Camilla selects the "Salesforce Example" connection
-    And she selects "Create Opportunity" integration action
-    And Camilla clicks on the "Done" button
+    When Camilla selects the "HTTP Example" connection
```
dsimansk: This is a good example, but I guess with should make it a bit better and specify correct properties in syndesis-rest jsondb. Because single URI is not valid anymore.

dsimansk:
Lets merge this manually as templates are broken now. Due to re-encrypt route issue...

PR #132, opened 2017-10-25
Adding scenario option into xvfb script

mcada:


ui-tests-protractor/e2e-xvfb.sh
```diff
@@ -3,4 +3,10 @@
 nohup /usr/bin/Xvfb :99 -ac -screen 0 1280x720x24 &
 export DISPLAY=:99
 yarn --no-progress
-exec yarn e2e:syndesis-qe
+
+if [[ "x${SCENARIO}" == "x" ]] ; then
```
dsimansk: @mcada anything wrong with condition like `[[ -z "${SCENARIO} "]]`? Nevertheless I like the inventive approach :)

mcada: You don't have to use google if you are not familiar with bash and see this expression so I think in non-bash testsuite it is better to use it like this.

dsimansk: I always google bash conditions, always :)

PR #136, opened 2017-10-26
[fix] Dropdown changed to select

lvydra:


dsimansk:
@lvydra rebase on master might be helpful to get rid of previous commit.

PR #151, opened 2017-11-01
[ui-tests] Login test and cleanup

mastepan:


ui-tests/src/main/java/properties/PropertiesLoader.java
```diff
@@ -0,0 +1,59 @@
+package properties;
```
dsimansk: Lets use more precise package, like `io.syndesis.qe.*`

ui-tests/src/main/java/properties/PropertiesLoader.java
```diff
@@ -0,0 +1,59 @@
...
+
+/**
+ * Created by mastepan on 10/30/17.
+ */
+public final class PropertiesLoader {
+
+	private PropertiesLoader() {
+	}
+
+	private static Properties properties = new Properties();
```
dsimansk: I'd like to see class like `TestConfiguration` that would define all property keys as constant and would expose methods to retrieve configuration without reusing keys.

ui-tests/src/test/resources/default.properties
```diff
@@ -0,0 +1,3 @@
+login.url=https:/syndesis.192.168.xxx.xxx.nip.io
+login.username=x
+login.password=x
```
dsimansk: Lets add prefix like, `syndesis.ui.url`, `syndesis.ui.username` to determine different properties for account credentials etc.

ui-tests/src/test/java/features/login.feature
```diff
@@ -0,0 +1,7 @@
+Feature: Login
+
+# doesn't work but looks like a test
+
+Scenario: Log in to syndesis
+	Given Camilla logs in on Openshift login page
```
dsimansk: Try to follow Protractor tests as closely as possible.

ui-tests/src/main/java/properties/Utils.java
```diff
@@ -0,0 +1,14 @@
...
+ */
+public class Utils {
+	public static void sleep(long millis) {
+		try {
+			Thread.sleep(millis);
+		} catch (InterruptedException e) {
+			e.printStackTrace();
+		}
+	}
+}
```
dsimansk: I would create a new module that would consist of util classes and also test configuration.

PR #162, opened 2017-11-08
chore: Plug ui tests to new config structure

dsimansk:


tplevko:
LGTM, will add @mastepan, so he will get familiar with changes in his code

PR #167, opened 2017-11-10
General ancestor for syndesis java page objects

lvydra:
WDYT?
Getting rid of former AppPage and moving general methods like findButton(), clickButton() etc on abstract ancestor class. Then this methods will be available on root element of every descendant page object.

### DIFF
--- a/ui-tests/src/test/java/io/syndesis/qe/pages/SyndesisPageObject.java
+++ b/ui-tests/src/test/java/io/syndesis/qe/pages/SyndesisPageObject.java
@@ -1,0 +1,81 @@
+package io.syndesis.qe.pages;
+
+import static com.codeborne.selenide.Condition.visible;
+import static com.codeborne.selenide.CollectionCondition.sizeGreaterThan;
+import static com.codeborne.selenide.WebDriverRunner.getWebDriver;
+
+import org.openqa.selenium.By;
+
+import com.codeborne.selenide.Condition;
+import com.codeborne.selenide.ElementsCollection;
+import com.codeborne.selenide.SelenideElement;
+
+public abstract class SyndesisPageObject {
+	
+	private static final class Link {
+		public static final By HOME = By.cssSelector("a.navbar-brand");
+	}
+
+	abstract SelenideElement getRootElement();
+
+	abstract boolean validate();
+
+	public String getCurrentUrl() {
+		return getWebDriver().getCurrentUrl();
+	}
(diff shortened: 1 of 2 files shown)


ui-tests/src/test/java/io/syndesis/qe/pages/SyndesisPageObject.java
```diff
@@ -0,0 +1,62 @@
...
+	void clickElementRandom(String elementClassName) {
+		ElementsCollection elements = this.getElementsByClassName(elementClassName);
+		int index = (int) Math.floor(Math.random() * elements.size());
+		elements.get(index).shouldBe(visible).click();
+	}
+
+	void selectOption(SelenideElement selectElement, String option) {
+		ElementsCollection optionElements = selectElement.findAll(By.tagName("option"));
+		optionElements = optionElements.filter(Condition.exactText(option));
+		optionElements.first().shouldBe(visible).click();
```
mcada: Add check if there is exactly one element in optionElements to avoid/throw error?

lvydra: Thank you and general idea?

lvydra:
No, the root element is first element of each page object, not the first element of page on which page object is. So we are really using it to define subgroup of elements.

mastepan:
That's exactly what I mean. Getting it doesn't change anything when selecting elements by id, css... but it can complicate life when using xpath. It usually won't if you don't use full path xpath /rootElement/.../yourElement.

lvydra:
@mastepan It can give us standardized way of checking presence of desired page objects on document and in opposition to you i think it could be quit useful when using css selectors.

lvydra:
@mastepan I think its good topic for office discussion.

Stefan365:
on nomenclature.
Some page objects we name "Component" (class name suffix) to show they are part of "Page" (suffix) objects. According to above structure is also Page object, but name Component suffix refers it's part of bigger page object. (example: ConnectionsDetailsComponent)
Do we keep this name distinction?

Stefan365:
is not SyndesisPageObject.validate() method obsolete, 
due to implementation SyndesisPageObject.getRootElement()
contains kind of validation (return $(Element.ROOT).shouldBe(visible);)
?

lvydra:
@Stefan365 Validation does not always have to be only about root element presence.

PR #169, opened 2017-11-13
Db utils

dsimansk:
This PR add simple method to execute sql command on sampledb for various verification tasks.

+ some changes to logging config :)


utilities/src/main/java/io/syndesis/qe/utils/DbUtils.java
```diff
@@ -0,0 +1,61 @@
...
+import java.util.Properties;
+
+import io.fabric8.kubernetes.api.model.Pod;
+import io.fabric8.kubernetes.client.LocalPortForward;
+import lombok.extern.slf4j.Slf4j;
+
+@Slf4j
+public class DbUtils {
+
+	private static DbUtils INSTANCE;
```
tplevko: checkstyle? :) Uppercase  just for constants.

PR #176, opened 2017-11-14
chore: Add BDD example to rest-tests

dsimansk:
This PR adds simple BDD test that for the given namespace tries to clean&deploy Syndesis.

Note that I've changed Cucumber dependencies to latest version. There was org-id change hence the bigger change in poms.

With proper test.properties PoC is runnable like:
`mvn clean test -P rest -Dtest=CucumberTestsRunner -Dcucumber.options="--tags @clean-deploy"`


tplevko:
POC LGTM

PR #178, opened 2017-11-15
chore: Add method DbUtils::executeUpdate

dsimansk:
It's not nice, but it will work for a while :)


dsimansk:
Closing in favour of more polished approach by @tplevko :)

PR #184, opened 2017-11-16
[ui-tests] Common steps - several steps added

mastepan:
I would let all unimplemented methods as they are until they are needed in a scenario.
I think those methods (like check integration title) should be implemented in the page objects they belong to. We can delete them once we are sure we don't need them or we don't need to track them anymore.

### DIFF
--- a/ui-tests/src/test/java/io/syndesis/qe/steps/CommonSteps.java
+++ b/ui-tests/src/test/java/io/syndesis/qe/steps/CommonSteps.java
@@ -1,17 +1,18 @@
 package io.syndesis.qe.steps;
 
+import static com.codeborne.selenide.Condition.visible;
 import static com.codeborne.selenide.Selenide.$;
 
 import org.assertj.core.api.Assertions;
 import org.openqa.selenium.By;
 import org.openqa.selenium.JavascriptExecutor;
 
-import com.codeborne.selenide.Condition;
 import com.codeborne.selenide.Selenide;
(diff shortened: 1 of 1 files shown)


ui-tests/src/test/java/io/syndesis/qe/steps/CommonSteps.java
```diff
@@ -67,6 +68,171 @@ public void resetState() {
...
+		syndesisRootPage.getRootElement().shouldBe(visible);
+	}
+
+	@Then("^(\\w+)? is presented with the \"([^\"]*)\" link*$/")
+	public void validateLink(String alias, String linkTitle) {
+		new SyndesisRootPage().getLink(linkTitle).shouldBe(visible);
+	}
+
+	@When("clicks? on the \"([^\"]*)\" button.*$/")
+	public void clickOnButton(String buttonTitle, String callback) {
```
dsimansk: We need to get rid off those `callback` params. Those aren't needed anymore. And why is there question mark in `@When("clicks? on the \"([^\"]*)\" button.*$/")`

mastepan: I Agree, the next iteration will fix this. I just wanted to add the methods as soon as possible without investigating how the regexp should look like. We can use them in our tests meanwhile.

dsimansk: So should I merge this PR then? Just say a word... :)

PR #200, opened 2017-11-23
[improve] Fill only present credentials

lvydra:


ui-tests/src/test/java/io/syndesis/qe/steps/connections/ConnectionSteps.java
```diff
@@ -86,8 +84,9 @@ public void typeConnectionDescription(String description) {
 	@When("^she fills \"([^\"]*)\" connection details$")
 	public void fillConnectionDetails(String connectionName) {
 		ConnectionConfigurationComponent connectionConfiguration = new ConnectionConfigurationComponent();
-		new AccountsDirectory().getAccount(connectionName).ifPresent(account -> connectionConfiguration.fillDetails(account.getProperties()));
-		//connectionConfiguration.fillDetails(this.world.getTestConfigConnection(connectionName));
+		String nameTransformed = connectionName.toLowerCase().replaceAll(" ", "_");
```
dsimansk: Lets get more heuristic approach, try to get with non-transformed name first and if result is empty, than tranform name. Wdyt?

PR #201, opened 2017-11-23
Login step updated

mcada:


ui-tests/src/test/java/io/syndesis/qe/steps/CommonSteps.java
```diff
@@ -48,8 +48,7 @@ public void login(String username) throws Throwable {
 
 		if (currentUrl.contains("api.fuse-ignite.openshift.com")) {
 			log.info("Ignite cluster login page");
-			$(By.className("login-redhat")).shouldBe(visible).click();
```
lvydra: What about something like that:
`SelenideElement login = $(By.className("login-redhat"));`
`if(login.isDisplayed()) {`
`login.click();`
`}`

PR #206, opened 2017-11-27
Connections feature migrated from protractor tests

mcada:


ui-tests/src/test/resources/features/connections.feature
```diff
@@ -0,0 +1,141 @@
+@smoke
```
dsimansk: The formatting is a bit weird. Could you just use 2 spaces instead of tabs?

ui-tests/src/test/java/io/syndesis/qe/steps/connections/ConnectionSteps.java
```diff
@@ -120,4 +120,10 @@ public void helpBlockVisible(String helpText) {
 		helpBlock.shouldBe(visible);
 		Assertions.assertThat(helpBlock.getText().equals(helpText)).isTrue();
 	}
+
+	@When("^clicks on the \"([^\"]*)\" kebab menu button of \"([^\"]*)\"$")
+	public void clicks_on_the_kebab_menu_button_of(String button, String connectionName) throws Throwable {
```
lvydra: @mcada Maybe it would be better to stick with java naming conventions and get rid of underscores. WDYT?

PR #210, opened 2017-11-29
fix: Fix selection of position in Integration editor

dsimansk:
I've noticed that `p.icon.active` is actually not nested inside `div.parent-active`. This code should be more precise in locating `STEP_ROOT` element and then look-up start of finish from it.


ui-tests/src/test/java/io/syndesis/qe/pages/integrations/edit/IntegrationFlowViewComponent.java
```diff
@@ -49,11 +54,11 @@ public String getIntegrationName() {
 	 *
 	 * @param type (start|finish)
 	 */
-	public FlowConnection flowConnection(String type) {
-		SelenideElement stepElement = this.getElementContainingText(Element.STEP, type).shouldBe(visible);
+	public FlowConnection flowConnection(final String type) {
```
lvydra: I'm not sure if we still need FlowConnection class, all we get is still only one element. If we just find element and return it, all will probably work same way. Only difference is isActive method which we don't use in current implementation.

Stefan365:
LGTM

ui-tests/src/test/java/io/syndesis/qe/pages/integrations/edit/IntegrationFlowViewComponent.java
Stefan365: @dsimansk just a note: then FlowConnection is redundant.

dsimansk: There is exactly one place that we use it :) So I guess I should remove the class and rework the isActive method to `IntegrationFlowViewComponent`?

https://github.com/syndesisio/syndesis-qe/blob/27b8df5c348e01756bdf168c7c0a2909207f657b/ui-tests/src/test/java/io/syndesis/qe/steps/integrations/IntegrationSteps.java#L297

dsimansk:
@lvydra @Stefan365 pls take a look again. I've removed the `FlowConnection` class and refactored method to be `verifyActivePosition`

Stefan365:
@dsimansk LGTM again

PR #211, opened 2017-11-30
Fix: wait for filters & wait for Active integration

dsimansk:


dsimansk:
@lvydra pls take a look at this one also.

PR #222, opened 2017-12-11
Rest update to use cucumber, verification steps moved to utils

tplevko:


rest-tests/src/test/resources/features/common.feature
```diff
@@ -1,7 +1,32 @@
...
+  @integrations-twitter-to-salesforce
+  Scenario: Tw - sf integration
+    Given user cleans tw to sf scenario
+    And user creates the tw connection using "twitter_listen" template
+    And user creates sf connection
+    And user creates tw mention step with "twitter-mention-connector" action
+    And user creates basic tw to sf filter step
+    And user creates tw to sf mapper step
+    And user creates sf step for tw sf test
+    And user creates tw to sf integration
```
dsimansk: I'd suggest to change this a bit a split this to 
```
When user creates tw to sf integration
Then wait for integration "<IntegrationName>" to be "Active"
```
wdyt?

rest-tests/src/test/resources/features/common.feature
```diff
@@ -1,7 +1,32 @@
...
+  Scenario: Clean & deploy
+    Given clean default namespace
+    When deploy Syndesis from template
+    Then wait for Syndesis to become ready
+
+  @integrations-twitter-to-salesforce
+  Scenario: Tw - sf integration
+    Given user cleans tw to sf scenario
+    And user creates the tw connection using "twitter_listen" template
+    And user creates sf connection
```
dsimansk: I don't see this step implementation, is it possible? :)

rest-tests/src/test/resources/features/common.feature
```diff
@@ -1,7 +1,32 @@
 Feature: Common scenarios
 
-@clean-deploy
-Scenario: Clean & deploy
-	Given clean default namespace
-	When deploy Syndesis from template
-	Then wait for Syndesis to become ready
\ No newline at end of file
+  @clean-deploy
```
dsimansk: Lets remove clean-deploy for now and stick just with `reset-db`

rest-tests/src/test/java/io/syndesis/qe/rest/tests/integrations/TwSfSteps.java
```diff
@@ -0,0 +1,124 @@
...
+	private final List<Step> steps = new ArrayList<>();
+	private AccountsDirectory accountsDirectory;
+
+	public TwSfSteps() throws GeneralSecurityException {
+		accountsDirectory = new AccountsDirectory();
+		connectorsEndpoint = new ConnectorsEndpoint(RestConstants.getInstance().getSyndesisURL());
+		connectionsEndpoint = new ConnectionsEndpoint(RestConstants.getInstance().getSyndesisURL());
+	}
+
+	@Given("^user creates tw mention step with \"([^\"]*)\" action")
```
dsimansk: Maybe we could have one generic method  like`user creates step from connector \"([^\"]*)\" with \"([^\"]*)\" action`

rest-tests/src/test/resources/features/common.feature
tplevko: all of the steps are implemented somewhere, it builds with no problem :)

rest-tests/src/test/resources/features/common.feature
tplevko: ok, sounds reasonable

rest-tests/src/test/resources/features/common.feature
tplevko: added, srry my bad

tplevko:
Updated. The generic steps you mentioned would be quite problematic to implement, there are many moving parts there, I will think about possible solution, but for now I think it is for the best to leave it as it is. The other things you highlighted should be reflected.

PR #230, opened 2017-12-12
fix: Remove annotation that breaks build

dsimansk:


dsimansk:
Lets close and reopen to see CI job triggered.

Issue #194, opened 2017-11-22
Page components relations overview


Stefan365:
In page-objects of ui-tests there are many page components.
For better overview snaphshots will be created to provide instant visual information about 'to which part of web-page component is related to'.
Also graphical map of relations between Classes would be of use for better understanding of system.

avano:
Closing all GH issues in syndesis-qe project, if it is still relevant, please open FUSEQE JIRA for that

PR #246, opened 2017-12-21
[relates #194] Readme

Stefan365:
readme for docs. first shot

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -1,0 +1,200 @@
+# Map of pages
+
+This document is create due to better orientation in the page components. 
+
+
+this documents speaks about:
+- 1. general architecture of our classes 
+- 2. identification of used elements
+
+## 1. General architecture
+
+Generally 'Page' or 'Component' suffix of java class provides information,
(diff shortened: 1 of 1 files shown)


docs/README.md
```diff
@@ -0,0 +1,200 @@
...
+## 1. General architecture
+
+Generally 'Page' or 'Component' suffix of java class provides information,
+that such class represents web element, defined in the Element.ROOT variable.
+If there are any sub-elements these are listed in private static classes, like 
+Input, Select, Button, or generally: Element
+Disadvantage of this approach is that such 'Pages' does not represent web 'pages' as user sees them.
+
+To partly remedy this disadvantage class 'SyndesisPage' has been created.
+This class represents web 'page' from user point of view.
```
lvydra: Only small remark. 
I'm not sure, but I think this could be little bit confusing in context of previous sentences. SyndesisPage class is only enumeration, which give you opportunity to get instance of page object (even so called "component") based on defined natural language alias, nothing more.

PR #252, opened 2018-01-09
fix: Various JMS scenario fixes

dsimansk:


ui-tests/src/test/java/io/syndesis/qe/pages/SyndesisPageObject.java
```diff
@@ -158,8 +158,14 @@ public SelenideElement getInputBySelector(String selector) {
 	public void fillInput(String inputId, String value) {
 		SelenideElement input = this.getElementById(inputId);
 		assertThat(input.getTagName(), is("input"));
-		input.shouldBe(visible).clear();
-		input.shouldBe(visible).sendKeys(value);
+		if (input.getAttribute("type").equals("checkbox")) {
+			if (Boolean.parseBoolean(value)  != input.isSelected()) {
```
lvydra: What about just `input.setSelected(Boolean.valueOf(value));`

dsimansk: Yep, that'll work just fine, thanks.

PR #264, opened 2018-01-16
fix: rest-tests port forward issues

tplevko:


utilities/src/main/java/io/syndesis/qe/utils/SampleDbConnectionManager.java
```diff
@@ -29,19 +29,20 @@ public static SampleDbConnectionManager getInstance() {
 	}
 
 	private SampleDbConnectionManager() {
-		localPortForward = TestUtils.createLocalPortForward("syndesis-db", 5432, 5432);
 	}
 
 	public static Connection getConnection() {
 		final Properties props = new Properties();
 		props.setProperty("user", "sampledb");
+		localPortForward = TestUtils.createLocalPortForward("syndesis-db", 5432, 5432);
```
dsimansk: I'd put the creation of port-forwad into `if {}` clause actually to prevent multiple invocation when connection is live already.

PR #265, opened 2018-01-16
Sf options test

tplevko:


utilities/src/main/java/io/syndesis/qe/validation/SfDbValidationSteps.java
```diff
@@ -134,4 +176,12 @@ private void deleteSalesforceLead(ForceApi salesforce, String firstName, String
 		final Optional<Lead> lead = queryResult.getTotalSize() > 0 ? Optional.of(queryResult.getRecords().get(0)) : Optional.empty();
 		return lead;
 	}
+
+	private Optional<Lead> getSalesforceLeadByEmail(ForceApi salesforce, String emailAddress) {
+		final QueryResult<Lead> queryResult = salesforce.query("SELECT Id,FirstName,LastName,Email,Company FROM lead where Email = '"
+				+ emailAddress + "'", Lead.class
+		);
+		final Optional<Lead> lead = queryResult.getTotalSize() > 0 ? Optional.of(queryResult.getRecords().get(0)) : Optional.empty();
```
dsimansk: What about `.orElse()`? 5th paragraph https://www.mkyong.com/java8/java-8-optional-in-depth/

pom.xml
```diff
@@ -35,6 +35,7 @@
 		<selenide.version>4.8</selenide.version>
 		<guava.version>22.0</guava.version>
 		<cucumber.version>2.1.0</cucumber.version>
+		<cucumber.spring>2.3.1</cucumber.spring>
```
dsimansk: We could probably upgrade all Cucumber deps with the `cucumber.version`, hopefully without issues. :)

tplevko: Ok, I was just afraid I would break something, but I think we could try it.

dsimansk: There're just 3 releases in between `2.1.0` vs. `2.3.1`. Nothing to worry about 🐉

utilities/src/main/java/io/syndesis/qe/validation/SfDbValidationSteps.java
tplevko: In this case I don't think it would be useful, at least as far as I investigated this proposed solution. The construction above prevents the IndexOutOfBounds that could be thrown by ...getRecords().get(0). I can't find a way to utilize the .orElse().

PR #266, opened 2018-01-16
Clean state via backend & verification in SF to DB

dsimansk:
A small disclaimer: I've deliberately kept `resetState()` via browser as a reference.

### DIFF
--- a/ui-tests/src/test/java/io/syndesis/qe/steps/CommonSteps.java
+++ b/ui-tests/src/test/java/io/syndesis/qe/steps/CommonSteps.java
@@ -125,7 +125,7 @@ public class CommonSteps {
 		allLinks.find(Condition.exactText(title)).shouldBe(visible).click();
 	}
 
-	@Given("^clean application state$")
+	//@Given("^clean application state$")
 	public void resetState() {
 		Long result = (Long) ((JavascriptExecutor) WebDriverRunner.getWebDriver())
 				.executeAsyncScript("var callback = arguments[arguments.length - 1]; " +
--- a/ui-tests/src/test/java/io/syndesis/qe/steps/CommonSteps.java
+++ b/ui-tests/src/test/java/io/syndesis/qe/steps/CommonSteps.java
@@ -125,15 +125,6 @@ public class CommonSteps {
 		allLinks.find(Condition.exactText(title)).shouldBe(visible).click();
 	}
 
-	//@Given("^clean application state$")
-	public void resetState() {
-		Long result = (Long) ((JavascriptExecutor) WebDriverRunner.getWebDriver())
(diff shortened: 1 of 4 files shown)


ui-tests/src/test/java/io/syndesis/qe/steps/CommonSteps.java
```diff
@@ -125,7 +125,7 @@ public void navigateTo(String username, String title) {
 		allLinks.find(Condition.exactText(title)).shouldBe(visible).click();
 	}
 
-	@Given("^clean application state$")
+	//@Given("^clean application state$")
 	public void resetState() {
```
tplevko: Can't this be removed completely?

utilities/src/main/java/io/syndesis/qe/bdd/CommonSteps.java
```diff
@@ -90,6 +91,14 @@ public void cleanTodoTable() {
...
 	}
 
+	@Given("^clean application state")
+	public void resetState() {
+		int responseCode = TestSupport.getInstance().resetDbWithResponse();
+		Assertions.assertThat(responseCode).isEqualTo(204);
+	}
+
+
+
```
tplevko: Why so many blank lines? :)

ui-tests/src/test/java/io/syndesis/qe/steps/CommonSteps.java
dsimansk: See the disclaimer :) I'd keep it a bit until new impl proof itself :)

utilities/src/main/java/io/syndesis/qe/bdd/CommonSteps.java
dsimansk: It was IDE not me, I swear...

utilities/src/main/java/io/syndesis/qe/bdd/CommonSteps.java
```diff
@@ -90,6 +91,9 @@ public void cleanTodoTable() {
 		SampleDbConnectionManager.getInstance().closeConnection();
 	}
 
-
-
+	@Given("^clean application state")
+	public void resetState() {
+		int responseCode = TestSupport.getInstance().resetDbWithResponse();
+		Assertions.assertThat(responseCode).isEqualTo(204);
```
dsimansk: @tplevko I've gone for checking response code with assert. 
But it might be worth to extend it a bit more in the future to add:
- retry few times if 204 is not returned
- fallback to browser reset
- assert the jsondb state directly.

tplevko: ok, sounds cool, will mark it as TODO for me

Issue #281, opened 2018-01-24
DB verifying steps to ui-test scenarios


Stefan365:
adding database verifying steps to ui-test scenarios where applicable.

dsimansk:
PRs merged, closing.

PR #285, opened 2018-01-30
[relates #281] DB verifying steps to scenarios. part 1

Stefan365:
DB veryfying steps, part 1

### DIFF
--- a/utilities/src/main/java/io/syndesis/qe/bdd/validation/DbValidationSteps.java
+++ b/utilities/src/main/java/io/syndesis/qe/bdd/validation/DbValidationSteps.java
@@ -4,9 +4,12 @@ import org.assertj.core.api.Assertions;
 
 import java.sql.ResultSet;
 import java.sql.SQLException;
+import java.util.ArrayList;
+import java.util.Iterator;
+import java.util.List;
 import java.util.concurrent.TimeUnit;
 
-import cucumber.api.PendingException;
+import cucumber.api.DataTable;
 import cucumber.api.java.en.Given;
 import cucumber.api.java.en.Then;
 import io.syndesis.qe.endpoints.TestSupport;
@@ -18,7 +21,7 @@ import lombok.extern.slf4j.Slf4j;
 
 /**
  * DB related validation steps.
- *
+ * <p>
  * Jan 17, 2018 Red Hat
  *
  * @author tplevko@redhat.com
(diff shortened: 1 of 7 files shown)


utilities/src/main/java/io/syndesis/qe/utils/DbUtils.java
```diff
@@ -75,4 +103,59 @@ public void deleteRecordsInTable(String tableName) {
 			log.error("Error: " + ex);
 		}
 	}
+
+	public Integer insertInDB(String tableName, List<List<String>> dataTable) {
```
dsimansk: I wonder if the same result could be achieve by `executeSqlOnSampleDb()` and processing the result outside DbUtil class. This class should stay pretty genetic and the testing logic executed in verifying steps.

cc @tplevko wdyt?

utilities/src/main/java/io/syndesis/qe/utils/DbUtils.java
```diff
@@ -75,4 +103,59 @@ public void deleteRecordsInTable(String tableName) {
...
+			int newId = st.executeUpdate();
+			log.info("***NEW INT: *{}*", newId);
+			return newId;
+		} catch (SQLException e) {
+			log.error("SQL ERROR: " + e);
+		}
+		return null;
+	}
+
+	public List<Integer> getCompletedByTask(String task) {
```
dsimansk: Same as for method above.

Stefan365: @dsimansk I have feeling construction of SQL query should be part of DbUtils, WDYT?

dsimansk: As I've described in the other comment. DbUtils should be more like entry point how to retrieve desired data for a test, not necessary hide whole validation logic etc. With a retrieved ResultSet you have more flexibility to assert all kind of values when needed.

ui-tests/src/test/java/io/syndesis/qe/pages/integrations/detail/IntegrationDetailPage.java
```diff
@@ -19,8 +19,8 @@
 	public final List<String> actionsSet = new ArrayList<String>();
 
 	private static final class Element {
-		public static final By ROOT = By.cssSelector("syndesis-integration-detail-page");
-		public static final By STATUS = By.cssSelector("syndesis-integration-status");
+		public static final By ROOT = By.cssSelector("syndesis-integrations-detail-page");
+		public static final By STATUS = By.cssSelector("syndesis-integrations-status");
```
lvydra: Are you sure with these? I'm not sure if we are using same version but i got integration without s.

Stefan365: in TP3 it is again with 's', (but not everywhere) see scenario @db-connection-crud-1-read-update comment

lvydra: OK, probably it will be best to wait.

dsimansk: I can see singular variant without `s`. I'd rather exclude it from PR for now and pls double-check the deployed version.

utilities/src/main/java/io/syndesis/qe/bdd/validation/DbValidationSteps.java
```diff
@@ -68,13 +74,53 @@ public void validateLead() {
 		Assertions.assertThat(getLeadTaskFromDb().toLowerCase()).isNotEmpty();
 	}
 
+	@Then("^validate add_lead procedure with last_name: \"([^\"]*)\", company: \"([^\"]*)\", period in ms: \"(\\w+)\"")
+	public void validateAddLeadProcedure(String lastName, String company, Integer ms) {
+		//wait for period cycle:
+		Selenide.sleep(ms + 100);
```
dsimansk: We definitely don't need Selenide's sleep, as it brings whole dependency. `Thread.sleep` or `TimeUnit.SECONDS.sleep` is good enough. If you mind the `InterruptedException` add something like sleepQuietly() into `TestUtils` class.

utilities/src/main/java/io/syndesis/qe/bdd/validation/DbValidationSteps.java
```diff
@@ -68,15 +71,96 @@ public void validateLead() {
...
+	@Then("^validate that all todos with task \"([^\"]*)\" have value completed \"(\\w+)\", period in ms: \"(\\w+)\"$")
+	public void checksThatAllTodosHaveCompletedVal(String task, Integer val, Integer ms) throws InterruptedException {
+		Thread.sleep(ms + 1000);
+
+		ResultSet rs;
+		List<Integer> completedAll = new ArrayList<>();
+		try {
+			String sql = String.format("SELECT completed FROM todo WHERE task like %s", task);
+			rs = dbUtils.readSqlOnSampleDb(sql);
+			while (rs.next()) {
```
dsimansk: I think this whole check can be simplified to 

```
Assertions.assertThat(rs).isEqualToComparingOnlyGivenFields(val, "completed");
```

utilities/src/main/java/io/syndesis/qe/bdd/validation/DbValidationSteps.java
```diff
@@ -68,15 +71,96 @@ public void validateLead() {
...
+				Integer compl = rs.getInt("completed");
+				completedAll.add(compl);
+			}
+			for (Integer compl : completedAll) {
+				Assertions.assertThat(compl).isEqualTo(val);
+			}
+		} catch (SQLException ex) {
+			log.error("Error: " + ex);
```
dsimansk: I'd also add Assertions.fail() upon SQL error.

ui-tests/src/test/java/io/syndesis/qe/pages/integrations/detail/IntegrationDetailPage.java
Stefan365: ok

utilities/src/main/java/io/syndesis/qe/bdd/validation/DbValidationSteps.java
Stefan365: ok

Issue #119, opened 2017-10-18
Create scenario to cover New Integration Lifecycle 


dsimansk:
@tplevko as discussed today pls continue with PoC scenario that will put to test switching between different states of Integration Deployment.

PR #295, opened 2018-02-07
[relates #119] Add basis for integration lifecycle testing

tplevko:


rest-tests/README.md
```diff
@@ -10,6 +10,20 @@ You may run the tests against a syndesis instance deployed on OpenShift as follo
...
 ```
 
+### Running parts of test suite
+
+There are cucumber tests deactivated by default. These tests contain tags namely:
+
+ * @integration-lifecycle
+ * @integration-lifecycle-long
+
+You can run them using following paramethers:
```
dsimansk: parameters

rest-tests/src/test/resources/features/integration-lifecycle.feature
```diff
@@ -0,0 +1,56 @@
...
+  Scenario: lifecycle integration act-deact
+    When create new integration with name: "DB to DB rest same name" and desiredState: "Active"
+	Then wait for integration with name: "DB to DB rest same name" to become active
+    Given create DB step with query: "SELECT * FROM TODO" and interval: 4000 miliseconds
+    And create DB insert taks step
+    Then try to create new integration with the same name: "DB to DB rest same name" and state: "Draft"
+
+  @integrations-lifecycle @integrations-lifecycle-long
+  Scenario: lifecycle integration act-deact 5 times
+    When create new integration with name: "DB to DB rest test act-deact" and desiredState: "Active"
```
dsimansk: This should be combination of 
```
When set...
The validate...
When set...
...
```

tplevko:
Fixes added

rest-tests/src/test/resources/features/integration-lifecycle.feature
dsimansk: Going through the second time. I've realized that the scenario could be implemented with a meta step that would encapsulate switching within a for-loop and take a param how many times. Wdyt?

tplevko: ok, will do

tplevko: Updated, have a look please.

utilities/src/main/java/io/syndesis/qe/bdd/validation/CommonValidationSteps.java
```diff
@@ -40,4 +44,68 @@ public void waitForIntegrationToBeActive(String integrationName) {
...
+			final Integration updatedIntegration = new Integration.Builder().createFrom(integration)
+					.desiredStatus(newDepState).build();
+			integrationsEndpoint.update(integration.getId().get(), updatedIntegration);
+
+			if (newDepState.name().equals("Active")) {
+				verifyPodCount(integrationName, 1);
+			} else {
+				verifyPodCount(integrationName, 0);
+			}
+			TestUtils.waitForState(integrationsEndpoint, updatedIntegration, newDepState, TimeUnit.MINUTES, 10);
```
dsimansk: Shouldn't there be an assert if the wait is successful?

tplevko: well, it should be indeed

PR #299, opened 2018-02-12
[ui-tests] Refactor IntegrationSteps

mastepan:
This refactor is not complete, there are more parts that are worth to revise later for better orientation in the testsuite.

### DIFF
--- a/ui-tests/src/test/java/io/syndesis/qe/pages/integrations/edit/IntegrationChooseStepComponent.java
+++ b/ui-tests/src/test/java/io/syndesis/qe/pages/integrations/edit/IntegrationChooseStepComponent.java
@@ -11,7 +11,7 @@ import io.syndesis.qe.pages.SyndesisPageObject;
 import lombok.extern.slf4j.Slf4j;
 
 @Slf4j
-public class IntegrationAddStepComponent extends SyndesisPageObject {
+public class IntegrationChooseStepComponent extends SyndesisPageObject {
 
 	private static final class Element {
 		public static final By ROOT = By.cssSelector("syndesis-integrations-step-select");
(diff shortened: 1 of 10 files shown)


ui-tests/src/test/java/io/syndesis/qe/pages/integrations/edit/IntegrationChooseStepComponent.java
```diff
@@ -11,7 +11,7 @@
 import lombok.extern.slf4j.Slf4j;
 
 @Slf4j
-public class IntegrationAddStepComponent extends SyndesisPageObject {
+public class IntegrationChooseStepComponent extends SyndesisPageObject {
```
lvydra: I would rather keep class names in align with root element names so `IntegrationStepSelectComponent` would be better in my opinion.

mastepan: The name of the element is "syndesis-integrations-step-select" which is different from the original class name IntegrationAddStepComponent... this is direct sign we should do something with class -naming... I have only aligned the name little bit more with the UI in this case. My original intention wasn't to smuggle in new class names in this PR ;)

lvydra: I know that this class have wrong name thats why I'm suggesting 'IntegrationStepSelectComponent' as name which is in align with our naming conventions.

mastepan: I have updated this as you wish but I still strongly disagree with the system. I need this PR to be merged to be ale to continue on my work. We can discuss the Page object class naming later...

lvydra: Its only suggestion. 
My point is that this is convention which we are using everywhere and we already discuss it. But I'm for another round of discussion.

PR #302, opened 2018-02-13
[utils] Springify the endpoints initialization

tplevko:


tplevko:
ok, should be good for a review, updated to match the new checkstyle

PR #303, opened 2018-02-13
Refactor api connector

mastepan:


ui-tests/src/test/java/io/syndesis/qe/steps/customizations/connectors/ApiClientConnectorsSteps.java
```diff
@@ -8,91 +8,41 @@
...
-                Assert.fail("The Auth type < " + authType + "> is not implemented by the test.");
-        }
+    @Then("^(\\w+) opens the API connector \"([^\"]*)\" detail$")
+    public void openApiConnectorDetail(String userName, String connectorName) {
+        apiClientConnectorsPage.clickConnectorByTitle(connectorName);
     }
 
     @Then("(\\w+) is presented with the new connector \"([^\"]*)\"$")
-    public void sheIsPresentedWithTheNewConnector(String user, String connectorName) throws Throwable {
+    public void checkNewConnectorIsPresent(String user, String connectorName) throws Throwable {
```
dsimansk: I wonder since you're on refactor path. Should be `user` parameter dropped also, as there's no use for it right now.

mastepan: It depends on the readability impact as we don't have roles in the app (yet?). We can discuss users in feature-steps separately. This applies to all scenarios.

dsimansk: No we don't yet. And imho we don't need to enforce role-to-account check on every step. I'd assume that once roles are present login method will take care of a user selection and we'd add a separate scenario to check if specific actions are reachable. E.g. start scenario with "login user with role xy" and all the steps performed in a sequence assume that actions are doable.

mastepan: Are we waiting for someone with merging?

dsimansk: Well, I guess we aren't. As we settled discussion around user on the meeting.

Issue #286, opened 2018-01-31
Investigate Monitoring Metrics feature


dsimansk:
Investigate progress made on https://github.com/syndesisio/syndesis/issues/1101 and if possible prepare PoC with simple tests to verify metrics are collected.

Stefan365:
there is still issue with metrics persistence for the time being, see my discussion with KurtStam here:
https://github.com/syndesisio/syndesis/issues/1101#issuecomment-365072425

dsimansk:
Done via #296 and #307. Further improvements should be added once UI is available.

PR #307, opened 2018-02-14
[relates #286] Rest metrics test. part 2

Stefan365:
some corrections


rest-tests/src/test/resources/features/integration-db-to-db-metrics.feature
```diff
@@ -0,0 +1,22 @@
+@integrations-db-to-db-metrics
+Feature: Test functionality of monitoring metrics system, part 1.
+
+  Background: Clean application state
+    And remove all records from DB
```
dsimansk: Shouldn't there be also general clean state?

Stefan365: done @ pushed again

Issue #287, opened 2018-01-31
Investigate Monitoring Logging feature


dsimansk:
Investigate progress made on https://github.com/syndesisio/syndesis/issues/1100 and prepare PoC with simpe tests if possible.

PR #313, opened 2018-02-19
[relates #287] Rest logs test

Stefan365:
activity/integrations log tests


utilities/src/main/java/io/syndesis/qe/endpoints/IntegrationsActivityEndpoint.java
```diff
@@ -0,0 +1,15 @@
...
+
+import java.security.GeneralSecurityException;
+
+import io.syndesis.model.metrics.IntegrationMetricsSummary;
+
+@Component
+public class IntegrationsActivityEndpoint extends AbstractEndpoint<IntegrationMetricsSummary> {
+
+    public IntegrationsActivityEndpoint() throws GeneralSecurityException {
+        super(IntegrationMetricsSummary.class, "/activity/integrations");
```
dsimansk: This doesn't seem right, when compared to `ActivityIntegrationsEndpoint` class.

PR #317, opened 2018-02-21
Support page version test updated

mcada:


ui-tests/src/test/java/io/syndesis/qe/pages/SupportPage.java
```diff
@@ -54,7 +58,8 @@ public File downloadZipLogs() throws InterruptedException {
     }
 
     public String getVersion() {
-        //TODO: lets wait here until they add ID for version string in the product
-        return "";
+        String versionString = getRootElement().$(Element.PRODUCT_VERSION).shouldBe(visible).getText();
+        log.info("I found this version string: " + versionString);
```
dsimansk: I'd prefer something less personified like `Version string found: {}`?

Issue #214, opened 2017-11-30
Refactor getElementContainingText method to target the search


dsimansk:
I'd like to be able to be more specific about where to look for and what to return. 

For example I want to get `integration-item` element from list. When I would use `getText()` on such an element the result would have name && status as a one string. 

Therefor I suggest that we introduce parent and text location element to be used if needed. 

@lvydra @Stefan365 wdyt?

```java
public SelenideElement getElementContainingText(By by, String text) {
		ElementsCollection elements = getRootElement().findAll(by).shouldBe(sizeGreaterThan(0));
		log.info("I found " + elements.size() + " elements.");

		elements = elements.filter(exactText(text));
		log.info("Elements after filter: " + elements.size());
		SelenideElement element = elements.shouldBe(sizeGreaterThan(0)).first();
		return element;
	}
```

Stefan365:
@dsimansk why not..

avano:
Closing all GH issues in syndesis-qe project, if it is still relevant, please open FUSEQE JIRA for that

PR #319, opened 2018-02-22
Twitter mention is now correctly followed by click on done button step

mcada:
Another change happened in master branch and when selecting Twitter mention we now have to click on done button.

### DIFF
--- a/ui-tests/src/test/java/io/syndesis/qe/pages/integrations/edit/ListActionsComponent.java
+++ b/ui-tests/src/test/java/io/syndesis/qe/pages/integrations/edit/ListActionsComponent.java
@@ -1,14 +1,12 @@
 package io.syndesis.qe.pages.integrations.edit;
 
-import static com.codeborne.selenide.Condition.visible;
-import static com.codeborne.selenide.Selenide.$;
-
-import org.openqa.selenium.By;
-
 import com.codeborne.selenide.SelenideElement;
-
 import io.syndesis.qe.pages.SyndesisPageObject;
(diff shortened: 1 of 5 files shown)


ui-tests/src/test/java/io/syndesis/qe/pages/integrations/edit/ListActionsComponent.java
```diff
@@ -33,5 +34,10 @@ public boolean validate() {
     public void selectAction(String name) {
         log.info("Searching for integration action {}", name);
         getElementContainingText(Element.NAME, name).shouldBe(visible).click();
+        if (name.equalsIgnoreCase("Mention")) {
+            //Can't use getButton as the root element is different :(
+            $$(By.tagName("button")).filter(Condition.matchText("(\\s*)" + "Done" + "(\\s*)"))
+                    .shouldHaveSize(1).first().click();
+        }
```
dsimansk: Imho this recalls my problem from https://github.com/syndesisio/syndesis-qe/issues/214. Fancy to refactor the whole method getElementContainingText? Is it the same case?

dsimansk: Actually this is super-specific fix, therefor I'd like to see more generic solution maybe :)

mcada:
Do not merge, I have to investigate another usage of this method in different context... Maybe we do not want to click "done" button at all... WIP

mcada:
Done.  I had to remove click in action method as in some cases we want to select an action and work with its options and not click straight away.

PR #329, opened 2018-03-02
Data buckets feature reworked due to changes in syndesis-ui

mcada:


dsimansk:
@mcada second commit message could be fixed maybe? :)

mcada:
Sorry, dashes :/

PR #334, opened 2018-03-06
chore: Add wait for all infra components

dsimansk:


utilities/src/main/java/io/syndesis/qe/bdd/CommonSteps.java
```diff
@@ -44,10 +50,29 @@ public void deploySyndesis() {
...
+                        .assertEventually();
+
+            executorService.submit(runnable);
+        });
+
+        executorService.shutdown();
+        try {
+            if (!executorService.awaitTermination(10, TimeUnit.MINUTES)) {
+                executorService.shutdownNow();
+                Assertions.fail("Syndesis didn't initilized in time");
```
avano: wasn't initialized

or

didn't initialize

utilities/src/main/java/io/syndesis/qe/bdd/CommonSteps.java
```diff
@@ -44,10 +50,29 @@ public void deploySyndesis() {
...
+            }
+        } catch (InterruptedException e) {
+            Assertions.fail("Syndesis didn't initilized in time");
```
avano: wasn't initialized

or

didn't initialize

PR #364, opened 2018-03-16
Data buckets tests extended

mcada:


ui-tests/src/test/resources/features/wip/data-buckets.feature
```diff
@@ -161,3 +162,215 @@ Feature: Test functionality of data buckets during creation of integration
 
 
 
+
```
dsimansk: Do you want to move this scenario from `wip` dir?

ui-tests/src/test/resources/features/wip/data-buckets.feature
```diff
@@ -161,3 +162,215 @@ Feature: Test functionality of data buckets during creation of integration
...
+    When she creates mapping from "firma" to "COMPANY"
+
+    And scroll "top" "right"
+    And click on the "Done" button
+
+
+
+    ##################### check warnings ##################################
+
+    And she checks that there is no warning inside of step number "0"
```
dsimansk: Could we refactor this step a bit to take a single number && range && list of steps to check maybe? Range is probably most valuable.

ui-tests/src/test/resources/features/wip/data-buckets.feature
```diff
@@ -161,3 +162,215 @@ Feature: Test functionality of data buckets during creation of integration
...
+    ##################### mappings ##################################
+    #################################################################
+
+    ##################### step step step step step MAPPING step ##################################
+
+    And She adds integration "step" on position "4"
+    And she selects "Data Mapper" integration step
+    Then she is presented with data mapper ui
+
+    Then check that data bucket "1 - SQL Result" is available and opens it
```
dsimansk: `Then check data bucket "xy" is available and open it` without a bit of grammar polish :)
 
But I'm thinking about a different approach like
```
When check data bucket "xy" is available
Then open data bucket "xy"
```
or 
```
Then check data bucket(s) and perform action(s)
  | Data Bucker name | state | action |
```

@mcada Wdyt?

ui-tests/src/test/resources/features/wip/data-buckets.feature
```diff
@@ -161,3 +162,215 @@ Feature: Test functionality of data buckets during creation of integration
...
+    When Camilla selects the "PostgresDB" connection
+    And she selects "Invoke SQL" integration action
+    Then she fills invoke query input with "insert into contact values (:#COMPANY , :#MYNAME , :#MYSURNAME , :#LEAD, '1999-01-01')" value
+    And clicks on the "Done" button
+
+    ##################### check warnings ##################################
+
+    And she checks that there is no warning inside of step number "0"
+
+    Then she checks that text "Add a data mapping step" is "visible" in step warning inside of step number "2"
```
dsimansk: Also a range here would be much nicer && clearer. Feel free to get rid off `she` subject. `The check text...` is sufficient imho.

ui-tests/src/test/resources/features/wip/data-buckets.feature
mcada: Data buckets are closed by default so when creating integration everything you need is this single step I created. Since this feature is testing if data bucket exists, no other feature will need just open/check - no point to test one thing more times. Close data bucket step may be useful, I am going to create it.

ui-tests/src/test/resources/features/wip/data-buckets.feature
mcada: This step will not be used outside of this feature. And even if so, it would be without range as steps you may want to check can be with odd/even or random indexes, depends if you already have data mapper or not. I think it is clear enough but if you insist I will create range for this specific case so it is just one row in feature file instead of 5.

ui-tests/src/test/resources/features/wip/data-buckets.feature
dsimansk: Ok, I'd like to tune down the repetitive step definition, that was the motication of step's table proposal(last one).

ui-tests/src/test/resources/features/wip/data-buckets.feature
dsimansk: @mcada sure it doesn't have to be range, but maybe an array of steps to check. You got me here I want to get rid off those lines...

Issue #367, opened 2018-03-19
Create AMQP connection e2e UI test


Stefan365:
UI test that covers AMQP connections.

dsimansk:
Done via linked PR.

PR #375, opened 2018-03-20
[relates #367] AMQP e2e UI test

Stefan365:
first shot.


utilities/src/main/java/io/syndesis/qe/bdd/CommonSteps.java
```diff
@@ -100,10 +100,15 @@ public void verifyBuild(String integrationName, int duration) {
     }
 
     @Given("^deploy AMQ broker$")
-    public void deployAMQBroker() throws Throwable {
+    public void deployAMQBroker() {
         AmqTemplate.deploy();
     }
 
+    @Given("^add \"([^\"]*)\" account$")
+    public void addAMQAccount(String protocol) {
```
dsimansk: Do we really need a dedicated step to add account? It could be a parameter of `deployAMQBroker`. But I'd prefer to keep it simple and just add both to our `AccountDirectory` and the scenario will just create required connection via `Given created connections` step.

Stefan365: @dsimansk I did it for a case with 2 different AMQ connections(openwire/amqp, e.g.) in the same integration , so that you need not to deploy amq server twice.

Stefan365: @dsimansk - done.

pom.xml
```diff
@@ -308,6 +309,14 @@
 				<artifactId>postgresql</artifactId>
 				<version>${postgresql.version}</version>
 			</dependency>
+
+			<!-- https://mvnrepository.com/artifact/org.apache.activemq/activemq-core 5.7.0 -->
+			<dependency>
+				<groupId>org.apache.activemq</groupId>
+				<artifactId>activemq-core</artifactId>
```
dsimansk: activemq-core is pretty old, it's better to use activemq-client in latest version probably.

utilities/src/main/java/io/syndesis/qe/utils/JmsUtils.java
```diff
@@ -0,0 +1,40 @@
...
+        jmsClient.sendMessage(msgText);
+    }
+
+    public String consumeMessage(String type, String destinationTo) throws Exception {
+        this.setClientDestination(type, destinationTo);
+        Message msg = jmsClient.receiveMessage();
+        if (msg instanceof TextMessage) {
+            return ((TextMessage) msg).getText();
+        } else {
+            throw new Exception("wrong type of message!!");
```
dsimansk: Without general Exception throw pls

ui-tests/src/test/java/io/syndesis/qe/steps/integrations/IntegrationSteps.java
```diff
@@ -136,7 +136,7 @@ public void expectIntegrationNotPresent(String name) {
     public void waitForIntegrationState(String integrationName, String integrationStatus) {
         SelenideElement integration = listPage.getListComponent().getIntegration(integrationName);
         assertTrue(TestUtils.waitForEvent(status -> status.equals(integrationStatus), () -> listPage.getListComponent().getIntegrationItemStatus(integration),
-                TimeUnit.MINUTES, 5, TimeUnit.SECONDS, 5));
+                TimeUnit.MINUTES, 1, TimeUnit.SECONDS, 1));
```
dsimansk: Any reason to change wait timeout to 1 min, it's too shot to get into Published?

PR #405, opened 2018-04-05
chore(rest): Adapt to new scenario names

dsimansk:
I've renamed the scenarios according our today's meeting.

In addition there are: 
- Annotated lifecycle scenario with `@wip` to skip it from execution now
- Merged activity and metrics scenario to one

//test: `@integration-1,@integration-2, @integration-3`


rest-tests/src/test/resources/features/integration-db-to-db-metrics.feature
```diff
@@ -1,27 +0,0 @@
...
-
-    And create finish DB invoke sql action step with query "INSERT INTO TODO (task, completed) VALUES (:#task, 0)"
-
-    Then create integration with name: "DB to DB metrics rest test"
-    Then wait for integration with name: "DB to DB metrics rest test" to become active
-
-    Then validate that number of all todos with task "Josef2" is greater than "1"
-
-#    monitoring metrics part:
-    Then validate that number of all messages through integration "DB to DB metrics rest test" is greater than "3", period in ms: "10000"
```
tplevko: Why did you remove this scenario? Have you discussed with Stefan?

dsimansk: Merged activity and metrics scenario to one

dsimansk: Those are basically same scenarios that has different last step. Imho is there any real values to keep them separately?

dsimansk: @tplevko but if you prefer to keep them separate I'll bring it back, np :)

tplevko: oh, didn't notice that, in that case it's ok

PR #428, opened 2018-04-17
[ui] Adding condition for visibility to all used elements

mcada:


ui-tests/src/test/java/io/syndesis/qe/fragments/common/list/RowList.java
```diff
@@ -32,12 +34,13 @@ public ElementsCollection getItemsCollection() {
 
     @Override
     public SelenideElement getItem(String title) {
-        return getTitle(title).$(By.xpath("./ancestor::*[@class='list-pf-item']"));
+        return getTitle(title).$(By.xpath("./ancestor::*[@class='list-pf-item']")).should(exist);
```
mastepan: If title exists/is visible, then the item it is in definitely exists too

PR #431, opened 2018-04-18
[ui-tests] Remove unused cucumber steps

mastepan:
I think we shouldn't have steps we don't use. Especially steps that check random subsets of the items.
If we really need them It's not so difficult to create them from existing deterministic steps. 

Well...or let's make a new task to create a scenario using these steps and I'll move them to the right *Steps class. Maybe I only put them to a wrong class. What do you think?


Stefan365:
I fully agree. if once needed we can find them in git history

Issue #445, opened 2018-04-25
integration e2e test with MySQL DB.


Stefan365:

tplevko:
Seems to be done now

PR #449, opened 2018-04-26
[relates #445] Mysql

Stefan365:
//test: `@integrations-db-to-db-mysql`

### DIFF
--- a/utilities/src/main/java/io/syndesis/qe/bdd/validation/DbValidationSteps.java
+++ b/utilities/src/main/java/io/syndesis/qe/bdd/validation/DbValidationSteps.java
@@ -33,7 +33,7 @@ public class DbValidationSteps {
     private final DbUtils dbUtils;
 
     public DbValidationSteps() {
-        dbUtils = new DbUtils(SampleDbConnectionManager.getConnection());
+        dbUtils = new DbUtils("postgresql");
     }
 
     @Given("^remove all records from table \"([^\"]*)\"")
@@ -84,8 +84,14 @@ public class DbValidationSteps {
                 2,
                 TimeUnit.SECONDS,
                 5);
-        Assertions.assertThat(contactCreated).as("Lead record has appeard in DB, todo table").isEqualTo(true);
+        Assertions.assertThat(contactCreated).as("Lead record has appeard in DB, TODO table").isEqualTo(true);
         Assertions.assertThat(getLeadTaskFromDb(lastName).contains(company));
+    }
+
+    @Then("^inserts into \"([^\"]*)\" table on \"([^\"]*)\"$")
+    public void insertsIntoTable(String tableName, String dbType, DataTable data) throws SQLException {
+        dbUtils.setConnection(dbType);
+        this.insertsIntoTable(tableName, data);
     }
 
     @Then("^inserts into \"([^\"]*)\" table$")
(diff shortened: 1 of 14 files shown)


utilities/src/main/java/io/syndesis/qe/bdd/validation/DbValidationSteps.java
```diff
@@ -30,10 +30,10 @@
 @Slf4j
 public class DbValidationSteps {
 
-    private final DbUtils dbUtils;
+    private DbUtils dbUtils;
 
     public DbValidationSteps() {
-        dbUtils = new DbUtils(SampleDbConnectionManager.getConnection());
```
mcada: Can we please maintain current idea about SampleDbConnectionManager having static method getConnection()?
The whole point was to not create db connection with every request/step and it was very well readable.
Your changes broke the pattern. DbUtils now need parameter to differentiate between mysql and postgre. What about leaving DbUtils as it is with connection parameter and change
``SampleDbConnectionManager.getConnection()``
to
``SampleDbConnectionManager.getConnection("databaseType")``

and implement whole logic inside of getConnection? I find your implementation chaotic.
One switch based on "databaseType" and one if-then-else per type would be enough.

With this approach you can also set it up that getConnection without data type will be default postgre to stay backward compatible.

Stefan365: just fyi, I don't create new DB connection on each request/step, but I generally, let's say your proposal is more clear...

Stefan365: done

utilities/src/main/java/io/syndesis/qe/utils/SampleDbConnectionManager.java
```diff
@@ -3,59 +3,113 @@
...
 public class SampleDbConnectionManager {
 
-    private static LocalPortForward localPortForward = null;
-    private static Connection dbConnection = null;
+    private static Map<String, DbWrapper> connectionsInfoMap = new HashMap<>();
 
+    public static Connection getConnection(String dbType) {
 
-    private SampleDbConnectionManager() {
+        DbWrapper wrap = new DbWrapper(dbType);
```
mcada: This is definitely better than previous version, just one last thing. You do not need to work with new wrap at all on this level.

getConnection(...) should be straight forward method like this:

1. try to get connection from static map connectionsInfoMap
2. check if it exists - yes? return it and exit. no? call method for connection creation like createConnection(dbType)
3. since you created it in previous step, now return connection from static map connectionsInfoMap

And in method createConnection(dbType) all you have to do is to create correct connection with help of switch and take advantage of static map parameter connectionsInfoMap.

The point is that you are sending instance of wrap class all over the place but in fact you do not need it at all - you only create it once and insert it into static map and then access to connector via that map.

dsimansk:
@Stefan365 could pls rebase to master to get latest CircleCI config changes.

Stefan365:
@dsimansk done & pushed

utilities/src/main/java/io/syndesis/qe/templates/MysqlTemplate.java
```diff
@@ -0,0 +1,134 @@
...
+                .withName("mysql-cmd")
+                .withContainerPort(3306)
+                .withProtocol("TCP").build());
+
+        Map<String, String> templateParams = new HashMap<>();
+        templateParams.put("MYSQL_USER", DB_USER);
+        templateParams.put("MYSQL_PASSWORD", DB_PASSWORD);
+        templateParams.put("MYSQL_DATABASE", DB_SCHEMA);
+
+        Collection<EnvVar> params = MysqlTemplate.repackAllParams(templateParams);
```
dsimansk: Why do we actually need templateParams in map first and then remap them into EnvVar list. Creating a list of EnvVar is not and option here?

PR #504, opened 2018-06-21
[FUSEQE-1175] Move logging configuration to util only to be shared be…

avano:
…tween modules


avano:
@mcada @dsimansk wdyt? this way the configuration is shared and we don't need 2 identical files in two modules :pregnant_woman:

PR #510, opened 2018-06-25
[ui][update] Improving version check in case of nightly build

mcada:


syndesisqe:
Build finished. Tests PASSED

PR #587, opened 2018-09-12
[fix] SlackConnector and Slack test was updated

mkralik3:
SlackConnector and Slack test was updated according to changed credentials.json.
Compatibility fix for 1.4.x.

//test: @slack-check-message-data-mapper


syndesisqe:

mkralik3:
updated

PR #594, opened 2018-09-18
Integration import export without salesforce

mmajerni:
Signed-off-by: mmajerni <mmajerni@redhat.com>
//test: `@integration-import-export`


syndesisqe:

mcada:
Next time please add "//test: @integration-import-export" into your PR description - it will run mentioned scenario, not only smoke tests. You can find additional information regarding PRs in readme file.

ui-tests/src/test/resources/features/integrations/integration-import-export.feature
```diff
@@ -5,7 +5,7 @@ Feature: Integration - Import Export
 
   Background: Clean application state
     Given clean application state
-    And reset content of "contact" table
+    Given reset content of "contact" table
```
mcada: "And" is correct step here - And has no meaning by itself - it inherits previous type of step. In this case the step is "Given". We did not respect this rule in the early developing phase but now we try to be more consistent with it so feel free to correct every scenario step you are working on. You can find more info about steps here: https://docs.cucumber.io/gherkin/reference/

mmajerni: Updated

mmajerni:
retest this please

mcada:
@mmajerni there is no need to ask for retest if you push new changes - it will be detected and tests will start automatically

syndesisqe:
Build finished. Tests FAILED

syndesisqe:
Build finished. Tests FAILED

mcada:
@mmajerni I fixed an issue why the second test failed, please rebase and stabilize the test - the first one failed.

syndesisqe:
Build finished. Tests FAILED

syndesisqe:
Build finished. Tests FAILED

syndesisqe:
Build finished. Tests FAILED

tplevko:
@mmajerni is there any progress on this PR?

syndesisqe:
Build finished. Tests FAILED

mcada:
retest this please

mcada:
last fail was caused by syndesis update while test was in progress :/

syndesisqe:
Build finished. Tests FAILED

syndesisqe:
Build finished. Tests FAILED

mmajerni:
retest this please

syndesisqe:
Build finished. Tests FAILED

syndesisqe:
Build finished. Tests PASSED

syndesisqe:
Build finished. Tests FAILED

tplevko:
retest this please

syndesisqe:
Build finished. Tests PASSED

ui-tests/src/test/resources/features/integrations/integration-import-export.feature
```diff
@@ -143,34 +134,33 @@ Feature: Integration - Import Export
...
 
-    When sleep for jenkins delay or "3" seconds
+    And sleep for jenkins delay or "3" seconds
+
+    Then check visibility of "Integration_import_export_test" integration details
     And navigate to the "Integrations" page
+
+    And Integration "Integration_import_export_test" is present in integrations list
+    # wait for integration to get in active state
+    And sleep for "300000" ms
```
tplevko: Please, for all the 5 min. and longer waits, find more dynamic solution. The whole test suite execution time would rise by more than 26 minutes and tbh I believe there is more elegant way to handle this.

ui-tests/src/test/resources/features/integrations/integration-import-export.feature
```diff
@@ -62,16 +54,14 @@ Feature: Integration - Import Export
     And click on the "Publish" button
     # assert integration is present in list
     Then check visibility of "Integration_import_export_test" integration details
+    And navigate to the "Integrations" page
 
-    When navigate to the "Integrations" page
-    Then Integration "Integration_import_export_test" is present in integrations list
+    And Integration "Integration_import_export_test" is present in integrations list
     # wait for integration to get in active state
+    And sleep for "300000" ms
```
mcada: This is not even imported integration so if the wait is necessary here, something is wrong with db to slack integration (not started in 10 mins = performance issue, please report it)

ui-tests/src/test/resources/features/integrations/integration-import-export.feature
```diff
@@ -95,13 +85,15 @@ Feature: Integration - Import Export
...
+    And Integration "Integration_import_export_test" is present in integrations list
+
+    And sleep for "300000" ms
     Then wait until integration "Integration_import_export_test" gets into "Running" state
 
-    # validate salesforce contacts
-    Then check that contact from SF with last name: "Jackson" has description "Red Hat"
-    # clean-up in salesforce
-    Then delete contact from SF with last name: "Jackson"
+    Then check that last slack message equals "Red Hat" on channel "test"
```
mcada: Isn't last message already "Red Hat" from previously run integration? maybe you should send something else to be sure that the message is new?

ui-tests/src/test/resources/features/integrations/integration-import-export.feature
```diff
@@ -95,13 +85,15 @@ Feature: Integration - Import Export
     And sleep for jenkins delay or "3" seconds
     # start integration and wait for published state
     And start integration "Integration_import_export_test"
+    Then check visibility of "Integration_import_export_test" integration details
+
     And navigate to the "Integrations" page
+    And Integration "Integration_import_export_test" is present in integrations list
+
+    And sleep for "300000" ms
```
mcada: Please try manually importing this integration to both minishift and openshift and tell us how long it takes to:
1. build the integration
2. deploying the integration
3. time from starting the integration till UI is informed that the integration is in running state

Do the same but create the integration from scratch instead of importing it and compare the results.

Create a notes document and keep it up to date with your findings - we may find some performance issues thanks to your work.

mmajerni: Here are tables with times:
[https://docs.google.com/spreadsheets/d/1I5FkFjHASaWzfeiI1tAqcyk33qqd2lzmHUALwew5yCc/edit?usp=sharing](url)

mkralik3:
In my point of view, the problem is that in the **test.properties** is set the timeout to 300s so every integration deployment which takes more than 5 minutes will fail. The timeout which is set [here](https://github.com/syndesisio/syndesis-qe/blob/master/ui-tests/src/test/java/io/syndesis/qe/steps/integrations/IntegrationSteps.java#L76) is ignored (or overridden).
Maybe you can create some method (_in the appropriate class_) which will update this [timeout](https://github.com/syndesisio/syndesis-qe/blob/master/ui-tests/src/test/java/io/syndesis/qe/CucumberTest.java#L20). And after the test, revert to the default value.
**e.g.**
```java
@Then("override timeout to \"(\\d+)\"s")
public void overrideTimeout(long second) {
   Configuration.timeout = second * 1000;
}

@Then("set default timeout")
public void setDefaultTimeout() {
    Configuration.timeout = TestConfiguration.getConfigTimeout() * 1000;
}
```
And after that call it in the test.
```cucumber
...

When navigate to the "Integrations" page
And override timeout to "400"s
Then wait until integration...
And set default timeout

...
```
It is the easy fix but the question is whether we want to do that this way.
@mcada @tplevko What do you think about it?

mcada:
@mkralik3 `Configuration.timeout` is just for selenide stuff, it should not influence other waiting utilities. Although it is true in this case - this step fails because method used in wait check is looking for web element which does not exist before integration is done with building and is in finished state - we should rework this part. No time out changer is required - maybe we should even consider lowering this timeout to 60s and rewrite parts which are using selenide methods incorrectly in wait utilities to lower total test run time in case of wrong element selectors.  @tplevko wdyt?

tplevko:
@mmajerni had you any chance to look at this please?

syndesisqe:
Build finished. Tests PASSED

syndesisqe:
Build finished. Tests FAILED

tplevko:
@mmajerni please investigate why the test failed and provide fix.

syndesisqe:
Build finished. Tests FAILED

tplevko:
@mmajerni seems the tests failed, could you have another look?

mmajerni:
retest this please

syndesisqe:
Build finished. Tests PASSED

PR #602, opened 2018-09-24
fix readme - how to run multiple scenarios via PR

mcada:
//skip-ci
@jstastny-cz thx for noticing, if you want to be proactive feel free to change the parsing and get rid of "@" symbol or create better logic ^^


syndesisqe:
Build finished. Tests PASSED

PR #621, opened 2018-11-02
Fix selection in Data Mapper

asmigala:
This was broken due to https://bugzilla.mozilla.org/show_bug.cgi?id=1421323
Changing CONTROL to META seems to work in both ff and chrome on linux.


syndesisqe:
Build finished. Tests PASSED

PR #628, opened 2018-11-14
[service-now] add ignore property for new parameters introduced in SN

mmelko:
add aditional check and cleanup incidents before test


syndesisqe:
Build finished. Tests PASSED

PR #647, opened 2018-11-23
[Operator] Undeploy syndesis before cleaning namespace

avano:
the xtf bump should resolve issue from @asmigala that sometimes the waitForSyndesis ends too soon

//skip-ci

### DIFF
--- a/README.md
+++ b/README.md
@@ -315,7 +315,9 @@ mvn clean test -P rest -Dcucumber.options="--tags @integration-ftp-ftp" \
 You can use various parameters. From the previous command, parameter:
 * *syndesis.config.openshift.namespace.lock* - lock the namespace
 * *syndesis.config.openshift.namespace.cleanup* - cleanup namespace before the tests
-* *syndesis.config.openshift.namespace.lock* - cleanup namespace after the tests (it can be useful to set this to false during debugging phase)
+* *syndesis.config.openshift.namespace.cleanup.after* - cleanup namespace after the tests (it can be useful to set this to false during debugging phase)
+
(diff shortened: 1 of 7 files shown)


syndesisqe:
Build finished. Tests FAILED

zregvart:
The intent is that the `OAuthClient` is removed when the Syndesis custom resource is removed, perhaps deleting it first and then the namespace would help? If that doesn't work then we're looking at a bug in the `syndesis-operator`.

syndesisqe:
Build finished. Tests FAILED

avano:
@zregvart 

currently we were deleting only selected resources https://github.com/xtf-cz/xtf/blob/utilities-0.6/src/main/java/cz/xtf/openshift/OpenShiftUtil.java#L822-L839 .

You are right, when I delete the custom resource, everything works as supposed. but since we want to clean the namespace is imho better to drop-create. 

wdyt? 

btw did I tickle your OAuth spidey-senses? :dancer:

zregvart:
> You are right, when I delete the custom resource, everything works as supposed. but since we want to clean the namespace is imho better to drop-create.
> 
> wdyt?

With using the operator we're also kinda testing that operator cleans up after itself. I'd like to move to a fully CR/operator driven thing for everything and make that as good experience as it can be. With all the shell/code we have outside of it we're just increasing our fragility. (2c) 

> btw did I tickle your OAuth spidey-senses? dancer

:spider: 

I'm kinda subscribed to everything and try to keep up, call me crazy...

syndesisqe:
Build finished. Tests PASSED

PR #685, opened 2019-01-14
[FUSEQE-2825] Create todo2 route after deploying syndesis

avano:
just a test if it helps, if not I'll reset it afterwards :-)


avano:
@mkralik3 it seems this helped - I changed the code to deploy the `todo2` route right after syndesis is deployed. Only thing that came to my mind why the previous approach didn't work was that it takes some time for the router to deploy the route (and in case of slow openstack it took more than 10 minutes - it is weird, but sadly thats openstack).

I don't know yet if it broke some other tests, do you think so?

mkralik3:
I am glad to hear. ;) thx, 
LGTM

PR #686, opened 2019-01-18
Allow to skip tests with open issues

asmigala:


syndesisqe:
Build finished. Tests PASSED

README.md
```diff
@@ -97,6 +97,10 @@ A scenario may optionally be tagged with one or more tags in the form `@gh-<issu
 If such a scenario fails, the relevant GitHub issues are checked and a summary is written in the cucumber report.
 See javadoc of the OnFailHooks class for more info.
 
+##### Skipping tests with open issues
+In some cases you might want to skip tests which you know are going to fail, because they are tagged with open issues.
+You can do that by setting the property `syndesis.skip.open.issues` to `true` (defaults to `false`).
```
avano: shouldn't this rather default to `true`? I think most of the time you want to skip them if they are open

avano:
@asmigala should the bogus feature be part of the PR?

README.md
asmigala: Hm, I was thinking, based on our discussion on the last meeting, that this is an "unsafe" feature - we want to be explicit that we are skipping some tests.
I was thinking we'd like to skip in nightly tests (where we just add this property to the jenkins job once and are done with it), but would not want to skip on product tests or locally (e.g. debugging a failed test locally would require adding this property).
Let me know what you think

asmigala:
> @asmigala should the bogus feature be part of the PR?

Of course not :D 
removed

syndesisqe:
Build finished. Tests PASSED

README.md
avano: well, both options are valid :-) We can leave it this way for now.

Will you add that property to jenkins job or should I do it?

PR #693, opened 2019-01-23
Change AMQ Message Broker to Red Hat AMQ

mkralik3:
//test: `@activemq`


ui-tests/src/test/resources/features/integrations/mqtt.feature
```diff
@@ -1,7 +1,6 @@
 # @sustainer: mcada@redhat.com
 
 @ui
-@activemq
```
avano: shoudln't this stay here?

mkralik3: I removed it because the test uses MQTT connector. I use @activemq tag only when a test uses AMQ connector (Red Hat AMQ connector now). e.g. @activemq tag is not in the ampq test either.
But it is true that those tests (mqtt, ampq) use AMQ broker.
So I can use @activemq tag for all tests where AMQ broker is present and @redhatamq for tests where the Red Hat AMQ connector is used.

What Do You Think?

avano: What about `@broker` for all the tests with broker and `@activemq` for the syndesis activemq connector?

mkralik3: LGTM

avano: may be

PR #717, opened 2019-02-14
[ui][fix] change save button name

mcada:
//skip-ci

### DIFF
--- a/ui-tests/src/test/resources/features/connections/oauth.feature
+++ b/ui-tests/src/test/resources/features/connections/oauth.feature
@@ -59,7 +59,7 @@ Feature: Connections - OAuth
     Then click on the "Done" button
 
     # finish and save integration
-    When click on the "Save as Draft" button
+    When click on the "Save" button
     And set integration name "Twitter to Salesforce oauth"
     And publish integration
     Then Integration "Twitter to Salesforce oauth" is present in integrations list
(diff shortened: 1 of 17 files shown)


ui-tests/src/test/resources/features/connections/oauth.feature
```diff
@@ -59,7 +59,7 @@ Feature: Connections - OAuth
     Then click on the "Done" button
 
     # finish and save integration
-    When click on the "Save as Draft" button
+    When click on the "Save" button
```
avano: can't this step be refactored to use this:

https://github.com/syndesisio/syndesis-qe/blob/04d6aee236532d1fb6d8c2317c069054507d363c/ui-tests/src/test/java/io/syndesis/qe/steps/integrations/editor/EditorSteps.java#L39

So that you don't have to change it in every test when the label changes?

mcada: Well yes, but there is a conflict between true BDD and programming best practices :/ I can have 150 steps with "click on X button" and reuse it whenever I want to click on that one button or I can find it dynamically and have just 1 step to find required button based on exact text. So either way there will be many steps or many methods. I prefer this way but if you insist I can refactor it.

avano: it depends on you :wink: you can have an enum of the button and find enum based on its name for example

mcada: Ofc there are ways, but button name from gherkin step should match what user sees anyway. Refactoring of gherkin steps is easy you can even use sed and replace whole step so in case of another change, group replace will be here.

avano: ¯\\_(ツ)_/¯

syndesisqe:
Build finished. Tests PASSED

PR #728, opened 2019-02-20
created sql connector reproducer test for gh-3856

jsafarik:
//test: `@sql-connector`


syndesisqe:

ui-tests/src/test/resources/features/integrations/sql-connector.feature
```diff
@@ -0,0 +1,64 @@
...
+    # save the integration with new name and publish it
+    When click on the "Save" button
+    And set integration name "DB_2_Log"
+    And publish integration
+    Then Integration "DB_2_Log" is present in integrations list
+    And wait until integration "DB_2_Log" gets into "Running" state
+
+    # validate that both items from db are present
+    When sleep for "10000" ms
+    Then .*validate that logs of integration "DB_2_Log" contains string "ABCsampleTODO"
```
asmigala: I believe this would pass even before - the problem happens when the SQL connection is in the middle of an integration

PR #744, opened 2019-03-01
[install] increase timeout when cleaning syndesis namespace

mcada:
//skip-ci


syndesisqe:
Build finished. Tests PASSED

avano:
why?

mcada:
Because it kept shutting down my syndesis local installation. There is hardcoded 20s timeout in used library and if all resources are not gone, it will stop the whole testsuite with one single fail.

mcada:
The same happened to @asmigala today more than once so we decided it needs to be fixed...

avano:
and isn't it because of the problems with the operator today?

mcada:
No, I encountered the issue before but couldn't reproduce it. Today it happened even with older syndesis tag.

avano:
@asmigala I'm just curious as I didn't encounter it yet :grin:

PR #745, opened 2019-03-01
[ui][fix] remove outdated step in concur feature

mcada:
//skip-ci

Skipping tests as they do not work for now - our concur credentials do not have correct permissions to access SAP concur instance endpoints, resulting in 403 error.


syndesisqe:
Build finished. Tests PASSED

PR #750, opened 2019-03-08
[FUSEQE-3430] Print pods when something bad happens

avano:
//skip-ci


syndesisqe:
Build finished. Tests PASSED

PR #817, opened 2019-03-27
fix api-client-connector

mmuzikar:
Added split steps before DB connections

ping @mastepan so we don't all fix the same tests


syndesisqe:
Build finished. Tests PASSED

PR #866, opened 2019-04-08
[ui] Skip apicurito standalone tests

asmigala:
These are included as part of the test-jar, but should not be run

//skip-ci


syndesisqe:
Build finished. Tests PASSED

PR #877, opened 2019-04-11
Add cucumber formatter for mail report

asmigala:
This is an initial implementation of showing open issues in emailed test report.

The basic idea is that there's a custom cucumber formatter that generates almost the whole report (except for some headers etc, which will need to be done exernally, since they need to know about the jenkins build info).

There are two main uglies in this:
1. We already check open issues in `IssueHooks.java`, so to prevent hitting github and zenhub twice for each failure, we embed the downloaded issues into the scenario and then retrieve it when creating the report. This does not seem to cause any trouble for the html report, but is not particularly nice. We might decide that requesting the issues twice is preferable
2. The report is currently built by appending to a `StringBuilder`, we might decide we want to use a templating engine to prevent the ugliness

There are also a bunch of todos which I will resolve once we decide this is the way we want to do things. The same goes for updating the jenkins job to use this new report.

@avano wdyt?

//skip-ci

### DIFF
--- a/utilities/src/main/java/io/syndesis/qe/cucumber/MailFormatter.java
+++ b/utilities/src/main/java/io/syndesis/qe/cucumber/MailFormatter.java
@@ -1,0 +1,183 @@
+package io.syndesis.qe.cucumber;
+
+import com.fasterxml.jackson.core.type.TypeReference;
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+import java.io.File;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+
+import cucumber.api.Result;
+import cucumber.api.TestCase;
+import cucumber.api.event.EmbedEvent;
+import cucumber.api.event.EventListener;
+import cucumber.api.event.EventPublisher;
+import cucumber.api.event.TestCaseFinished;
+import cucumber.api.event.TestRunFinished;
+import cucumber.api.event.TestSourceRead;
+import gherkin.AstBuilder;
+import gherkin.Parser;
+import gherkin.ParserException;
+import gherkin.ast.Comment;
+import gherkin.ast.Feature;
+import gherkin.ast.GherkinDocument;
+import io.syndesis.qe.issue.IssueState;
+import io.syndesis.qe.issue.SimpleIssue;
+import lombok.Data;
+import lombok.RequiredArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
+
+@Slf4j
+public class MailFormatter implements EventListener {
+
+    private static final Pattern SUSTAINER_PATTERN = Pattern.compile(".*@sustainer: ([a-zA-Z@.]+)");
+
+    private final String path;
+
+    private ObjectMapper mapper = new ObjectMapper();
+
+    private Map<String, String> sustainers = new HashMap<>();
+    private Map<String, Feature> features = new HashMap<>();
+    private Map<String, List<SimpleIssue>> scenarioIssues = new HashMap<>();
+    private Set<String> recipients = new HashSet<>();
+    private List<ScenarioResult> results = new ArrayList<>();
+
+    public MailFormatter(String path) {
+        this.path = path;
+    }
+
+    @Override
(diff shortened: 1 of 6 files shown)


syndesisqe:
Build finished. Tests PASSED

utilities/src/main/java/io/syndesis/qe/cucumber/MailFormatter.java
```diff
@@ -0,0 +1,145 @@
...
+
+    @Override
+    public void setEventPublisher(EventPublisher publisher) {
+        publisher.registerHandlerFor(TestSourceRead.class, this::onTestSourceRead);
+        publisher.registerHandlerFor(TestCaseFinished.class, this::onTestCaseFinished);
+        publisher.registerHandlerFor(TestRunFinished.class, this::onTestRunFinished);
+        publisher.registerHandlerFor(EmbedEvent.class, this::onEmbed);
+    }
+
+    private void onEmbed(EmbedEvent t) {
```
asmigala: There's a bug here that test which passed after a failed test will have that failed test's issues bound to them, will fix

syndesisqe:
Build finished. Tests PASSED

avano:
@asmigala superb stuff, saves a lot of bash-fu we do at the moment to parse the reports. Templating engine could be nice, but I'm also fine with current implementation

utilities/src/main/java/io/syndesis/qe/hooks/IssueHooks.java
```diff
@@ -37,7 +40,46 @@
 @Slf4j
 public class IssueHooks {
 
-    enum IssueState {
+    public static class SimpleIssue {
```
avano: you can use

```
@Getter
@Setter
@NoArgsConstructor
@AllArgsConstructor
```

utilities/src/main/java/io/syndesis/qe/cucumber/MailFormatter.java
```diff
@@ -0,0 +1,146 @@
...
+            // not much we could do
+        }
+    }
+
+    private GherkinDocument parseGherkinSource(String source) {
+        Parser<GherkinDocument> parser = new Parser<GherkinDocument>(new AstBuilder());
+        try {
+            return parser.parse(source);
+        } catch (ParserException e) {
+            // TODO: decide how to handle this error
```
avano: do you know when this may happen?

I think the only option is to log it and ¯\\_(ツ)_/¯ (and handle that it won't fail later on NPE)

avano:
@asmigala one thing that came to my mind - would it be worth to create a class representation of the report row and save the instances in some list and then just create the report by appending the `toString()` of the instances?

it's basically the same but you can build the report at once and not appending one row in `onTestCaseFinished`

utilities/src/main/java/io/syndesis/qe/hooks/IssueHooks.java
asmigala: Cool, didn't know about that, will change. Also, maybe I should refactor this outside of `IssueHooks`, wdyt?

utilities/src/main/java/io/syndesis/qe/cucumber/MailFormatter.java
asmigala: agreed. I think this should not really happen ever, since any parsing errors would prevent the tests from running

utilities/src/main/java/io/syndesis/qe/hooks/IssueHooks.java
avano: or theoretically 

```
@Data
@AllArgsConstructor
```

may be enough - now I'm wondering what the [@RequiredArgsConstructor](https://projectlombok.org/features/constructor) will do when there are no final/@NonNull fields - will it create `@NoArgsConstructor`? :thinking: 

anyway yes, a separate class would be nice

asmigala:
@avano sounds good, I'll do that.

One more thing, in the pipeline library, we currently keep a list of sustainers who's tests failed and then include them in the `To` field in the email. Now with this change, we could either post-process each line in the generated html report (which seems fragile to me), or we could generate the list of sustainers in the `MailFormatter` and put it in the report in some easy to parse format. We could even put just the list of maintainers in a separate file.

avano:
@asmigala +1 to separate file with space delimited mails

avano:
*comma-separated

syndesisqe:
Build finished. Tests PASSED

syndesisqe:
Build finished. Tests PASSED

avano:
@asmigala great stuff, I assume there should be backport/master label as well :yum:

PR #893, opened 2019-04-24
[ui][fix] fail the validation step when result set thorws an exception

mcada:
//test: `@db-connection-crud-1-read-update`


syndesisqe:
Build finished. Tests PASSED

PR #894, opened 2019-04-25
[ui][fix] fail the validation step when result set thorws an exception

mcada:
//skip-ci


syndesisqe:
Build finished. Tests PASSED

PR #906, opened 2019-05-13
[Connections] Refactor creating connections

avano:


syndesisqe:
Build finished. Tests PASSED

rest-tests/src/test/java/io/syndesis/qe/rest/tests/steps/connection/Connections.java
```diff
@@ -34,6 +34,7 @@
     @Autowired
     private ConnectorsEndpoint connectorsEndpoint;
     private final AccountsDirectory accountsDirectory;
+    private Account account;
```
asmigala: I'm a bit torn about this - I understand that we store the account as an instance field to avoid repeating it in the `accountProperty` calls, but it seems kind of hacky.

avano: okay :(
okay :(

PR #909, opened 2019-05-16
[ui][new] allow to set webdriver version from maven command to support older browsers

mcada:
//skip-ci

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -60,6 +60,16 @@
 		<box.version>2.30.1</box.version>
 		<jsoup.version>1.11.3</jsoup.version>
 		<aws.sqs.version>2.5.29</aws.sqs.version>
+
+		<!-- leave driver versions empty to use latest driver -->
+		<!-- Find supported webdriver version for your chrome browser here:
+		 	http://chromedriver.chromium.org/downloads -->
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -59,6 +59,8 @@
 		<el.version>3.0.0</el.version>
 		<box.version>2.30.1</box.version>
 		<jsoup.version>1.11.3</jsoup.version>
+		<chrome.driver.version></chrome.driver.version>
+		<firefox.driver.version></firefox.driver.version>
```
mcada: I can delete it from here or leave it empty so it is visible that the property exists from the main pom.xml, what do you suggest @avano ?

syndesisqe:
Build finished. Tests PASSED

pom.xml
avano: put a comment there that empty == latest

avano:
@mcada what does the version look like? it is just `73` or something else?

pom.xml
asmigala: I suggest adding it to the readme as well

syndesisqe:
Build finished. Tests PASSED

syndesisqe:
Build finished. Tests PASSED

PR #911, opened 2019-05-17
[Drivers] Bind drivers download to install phase

avano:
so that they are not re-downloaded when using `mvn test`

//skip-ci


syndesisqe:
Build finished. Tests PASSED

PR #914, opened 2019-05-20
Reproducer for gh-4466: SQL connector should return generated keys

jsafarik:
//test: `@sql-connector-return-keys`

### DIFF
--- a/ui-tests/src/test/resources/features/integrations/sql-connector.feature
+++ b/ui-tests/src/test/resources/features/integrations/sql-connector.feature
@@ -249,3 +249,35 @@ Feature: SQL Connector
     # Wait for "Running" state and validate
     And wait until integration "Generated_Keys" gets into "Running" state
     And validate that logs of integration "Generated_Keys" contains items with IDs "1"
+
+  @reproducer
+  @gh-5493
+  @sql-connector-return-keys-start-step
+  Scenario: Return generated keys for INSERT statement from start step
+
+    When click on the "Create Integration" button to create a new integration
+    Then check visibility of visual integration editor
+    And check that position of connection to fill is "Start"
+
+    # Start step
+    And select the "PostgresDB" connection
+    And select "Periodic SQL invocation" integration action
+    And fill in invoke query input with "INSERT INTO todo (task, completed) VALUES ('task1', 0)" value
+    And click on the "Next" button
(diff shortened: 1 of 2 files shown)


syndesisqe:
Build finished. Tests PASSED

ui-tests/src/test/resources/features/integrations/sql-connector.feature
```diff
@@ -203,8 +203,61 @@ Feature: SQL Connector
...
+      | task | task |
+    And click on the "Done" button
+    And publish integration
+    And set integration name "Generated_Keys"
+    And click on the "Publish" button
+
+    # Wait for "Running" state and validate
+    And wait until integration "Generated_Keys" gets into "Running" state
+    Then validate that number of all todos with task "task1" is greater than "1"
+    And validate that logs of integration "Generated_Keys" contains bodies of items "2"
```
asmigala: I would simplify this:

1. Periodic invocation with `INSERT INTO todo (task) VALUES ('whatever')`
2. Log step 

then you can just check that you get `{"id": 1}` in the log. No need to select and reinsert the data.

Also, after discussing with @mcada , we'd like to rename the validation step to 
`validate that logs of integration "<integration>" contain items with IDs "<ids>"` to make it clearer, could you please also include this in you PR?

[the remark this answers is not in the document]
jsafarik: Done

jsafarik: So I've simplified it to Timer -> SQL insert -> Log (since having the SQL insert step as start step doesn't return keys right now, issue [#5493](https://github.com/syndesisio/syndesis/issues/5493) is reporting this).
Also, renamed the verification step as requested.

syndesisqe:
Build finished. Tests PASSED

asmigala:
Could please also include a reproducer for https://github.com/syndesisio/syndesis/issues/5493 ? Should be almost the same as this one a since you are already touching this, it seems more efficient.

jsafarik:
@asmigala 
So I've added it to this PR as well, but the added test is failing since it is not fixed yet. but since this PR runs only the first test, it should pass.

syndesisqe:
Build finished. Tests PASSED

PR #929, opened 2019-05-27
[checkstyle] Tweak checkstyle, prevent bypassing by not staging

asmigala:
//skip-ci

### DIFF
--- a/hooks/prepare-commit-msg
+++ b/hooks/prepare-commit-msg
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 
 CHECKSTYLE="checkstyle-8.20"
 CHECKSTYLE_JAR="${CHECKSTYLE}-all.jar"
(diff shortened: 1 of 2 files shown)


syndesisqe:
Build finished. Tests PASSED

hooks/prepare-commit-msg
```diff
@@ -13,9 +13,19 @@ COMMIT_MSG="$(cat "$1")"
 
 [ -z "${FILES}" ] && exit 0
 
+TMP_DIR="$(mktemp -d)"
+trap "rm -rf $TMP_DIR" 0
+while read -r F ; do 
+    FILE_DIR="$(dirname "$F")"
+    mkdir -p "$TMP_DIR/$FILE_DIR"
+    git show ":$F" > "$TMP_DIR/$F"
+done <<< "$FILES"
```
avano: what about:

```
trap "git stash pop" EXIT
git stash save -q
checkstyle
```
?

avano: you know, with some check that stash saved something :grin:

asmigala: What would be the benefit of that? I prefer not to touch the working dir if possible

avano: it seemed to me less hacky :-)

anyway, this one fails for me:

```
gc -m "test"                                        
.git/hooks/prepare-commit-msg: 22: .git/hooks/prepare-commit-msg: Syntax error: redirection unexpected
```

asmigala: Ah, probably because of [this](https://stackoverflow.com/a/2462357), are you ok with changing the hashbang to `bash`, or do you want me to go the `git stash` route?

avano: one very very minor issue is that it will print the path to the temp dir in the checkstyle:

```
Starting audit...
[ERROR] /tmp/tmp.yec7ynYjvZ/utilities/src/main/java/io/syndesis/qe/Component.java:19: 'member def type' has incorrect indentation level 3, expected level should be 4. [Indentation]
Audit done.
Checkstyle ends with 1 errors.
```

but that will probably hurt only my eyes :grin: so you can change it to bash and if someone else will complain, then we can refactor it

asmigala: Should be easier on your eyes now

syndesisqe:
Build finished. Tests PASSED

hooks/prepare-commit-msg
avano: @asmigala I promise this is a last nitpick:
since we have $ROOT_DIR and $TMP_DIR, what about something like that?

```
OUT=$(java -jar "${ROOT_DIR}/checkstyle/${CHECKSTYLE_JAR}" -c "${ROOT_DIR}/checkstyle/checkstyle.xml" ${FILES})
EXIT=$?
echo "${OUT}" | sed "s#${TMP_DIR}#${ROOT_DIR}#g"
exit ${EXIT}
```

asmigala: Hm, I can add that, but wouldn't that be confusing? Like, you make changes, checkstyle complains, so you fix them but forget to `git add` them, checkstyle complains again and tells you which file you need to fix, you open it and the fixed changes are there! What the hell, checkstyle?
When you see the tmp dir in the log, you are much more likely to realize that something is going on.

avano: I can imagine the first problem will rather be "I did the changes in my repository but the checkstyle probably uses some old version of the file in some random temp directory"

asmigala: Hm, we could also indicate to the user that they have unstaged files, but not sure that should be the reponsibility of this hook

avano: I think that is not necessary. For me personally, if the checkstyle fails because I forgot to add files, error with git repo path tells me that something  is probably wrong with my git repository

while error with tmp:

asmigala: Ok, but you'll be the one explaining to others why checkstyle doesn't see their changes :D

syndesisqe:
Build finished. Tests PASSED

hooks/prepare-commit-msg
avano: I'm ready :D

PR #930, opened 2019-05-28
[FUSEQE-4498] fix tests with sql connector mapping collections

wyvie:
I returned the `DataMapper#openDataMapperCollectionElement` back into `create data mapper mappings` step so that any collections we can find on the data mapping view are opened automatically when we try to create the mapping.

It was previously like that before and was removed for the unknown reason, so I put it back to fix the UI tests with SQL connector collections problems.

Here it also checks that collection is not yet open by looking only for `.fa-angle-right` arrow and clicking it. Opened collections have `.fa-angle-down` arrow.


syndesisqe:
Build finished. Tests FAILED

syndesisqe:
Build finished. Tests PASSED

PR #932, opened 2019-05-28
[ui] - update activity step test according to fix #4181 

mkralik3:
update activity step test according to fix #4181 and skip error test due to #4192

//test: `@activity-test`


syndesisqe:
Build finished. Tests PASSED

PR #933, opened 2019-05-28
Datamapper - fix null id collection mapping

mastepan:
//test: `@gmail-receive`

### DIFF
--- a/ui-tests/src/test/java/io/syndesis/qe/pages/integrations/editor/add/steps/DataMapper.java
+++ b/ui-tests/src/test/java/io/syndesis/qe/pages/integrations/editor/add/steps/DataMapper.java
@@ -56,6 +56,7 @@ public class DataMapper extends SyndesisPageObject {
         public static final By PRIMARY_BUTTON = By.cssSelector("button.btn-primary");
         public static final By MAPPER_COLLECTION_ICON = By.className("parentField");
         public static final By ADD_MAPPING_ICON = By.className("fa-plus");
+        public static final By NULL_ID_COLLECTION_CONTAINER = By.cssSelector("\"div.DocumentFieldDetailComponent\"");
     }
 
     @Override
@@ -174,14 +175,47 @@ public class DataMapper extends SyndesisPageObject {
         List<String> path = Arrays.asList(mappingName.trim().split("\\."));
         final String parent = path.get(0);
 
-        path.forEach(s -> {
-            SelenideElement detailElement = containerElement.find(By.id(s)).shouldBe(visible);
+        boolean previousHadNullId = false;
+
+        for( int i = 0; i < path.size(); i++ ) {
+            String s = path.get(i);
+
+            //id is null
+            if (s.equals("")) {
+                previousHadNullId = true;
+                continue;
+            }
+
+            SelenideElement detailElement = containerElement.find(By.id(s));
+
+            //previous id was null
+            if (previousHadNullId) {
+                if (!detailElement.is(visible)) {
+                    containerElement.find(Element.NULL_ID_COLLECTION_CONTAINER).shouldBe(visible).click();
+                }
+            }
+
+            if (detailElement.exists()) {
+                detailElement.shouldBe(visible);
+            } else {
+                containerElement.find(Element.NULL_ID_COLLECTION_CONTAINER).shouldBe(visible).click();
+                detailElement.shouldBe(visible);
+            }
+
             if (detailElement.find(Element.CHILDREN).exists()) {
                 // if there're childrenFields display element is expanded already, click otherwise
             } else {
                 SelenideElement el;
                 if (!s.equals(parent)) {
-                    el = containerElement.find(By.id(parent)).find(By.id(s)).$(Element.LABEL).shouldBe(visible);
+                    By parentElementSelector = null;
+
+                    if(previousHadNullId || i == 0) {
+                        parentElementSelector = Element.NULL_ID_COLLECTION_CONTAINER;
+
+                    } else {
+                        parentElementSelector = By.id(parent);
+                    }
+                    el = containerElement.find(parentElementSelector).find(By.id(s)).$(Element.LABEL).shouldBe(visible);
                 } else {
                     el = detailElement.$(Element.LABEL).shouldBe(visible);
                 }
@@ -194,10 +228,10 @@ public class DataMapper extends SyndesisPageObject {
                         .click()
                         .keyUp(Keys.META)
                         .perform();
-
             }
 
-        });
+            previousHadNullId = false;
+        }
     }
 
     /**


mcada:
@mastepan please add a test annotation which was failing due to this bug

mastepan:
@mcada I know, thanks. I need to find out what tests are failing first. I did id for my new not-yet-merged tests but it may help others too.

syndesisqe:
Build finished. Tests PASSED

ui-tests/src/test/java/io/syndesis/qe/pages/integrations/editor/add/steps/DataMapper.java
```diff
@@ -174,14 +174,50 @@ public void selectMapping(String mappingName, SelenideElement containerElement)
...
+                previousHadNullId = true;
+                continue;
+            } else {
+                if (previousHadNullId) {
+                    log.info("WWWWW previous ID was null");
+                    log.info("WWWWW current  ID is  " + s);
+
+                    if (!detailElement.is(visible)) {
+                        log.info("WWWWW clicking an ancestor");
+                        containerElement.find(By.cssSelector("div[class=\"DocumentFieldDetailComponent\"")).shouldBe(visible).click();
```
asmigala: this css selector is used twice, please refactor it into the `Element` class.

Also, any particular reason to use `div[class="DocumentFieldDetailComponent"]` rather than `div.DocumentFieldDetailComponent`? Should work the same unless I'm missing something

ui-tests/src/test/java/io/syndesis/qe/pages/integrations/editor/add/steps/DataMapper.java
```diff
@@ -174,14 +174,50 @@ public void selectMapping(String mappingName, SelenideElement containerElement)
...
+                    } else {
```
asmigala: remove unnecessary `else` branch

ui-tests/src/test/java/io/syndesis/qe/pages/integrations/editor/add/steps/DataMapper.java
```diff
@@ -174,14 +174,50 @@ public void selectMapping(String mappingName, SelenideElement containerElement)
...
+
+        for( int i = 0; i < path.size(); i++ ) {
+            String s = path.get(i);
+            SelenideElement detailElement = containerElement.find(By.id(s));
+
+            if (s == null || s.equals("")) {
+                log.info("WWWWW ID is null");
+                previousHadNullId = true;
+                continue;
+            } else {
```
asmigala: Since we are `continue`ing in the `if` above, this `else` branch could be removed and the contents moved out directly into the `for`. That's more of a style issue than anything else, @mcada wdyt?

ui-tests/src/test/java/io/syndesis/qe/pages/integrations/editor/add/steps/DataMapper.java
```diff
@@ -174,14 +174,50 @@ public void selectMapping(String mappingName, SelenideElement containerElement)
...
-        path.forEach(s -> {
-            SelenideElement detailElement = containerElement.find(By.id(s)).shouldBe(visible);
+        boolean previousHadNullId = false;
+
+        for( int i = 0; i < path.size(); i++ ) {
+            String s = path.get(i);
+            SelenideElement detailElement = containerElement.find(By.id(s));
+
+            if (s == null || s.equals("")) {
+                log.info("WWWWW ID is null");
```
asmigala: What is the purpose of the `WWWW` in the log? I assume it's for debugging, please remove.

ui-tests/src/test/java/io/syndesis/qe/pages/integrations/editor/add/steps/DataMapper.java
```diff
@@ -174,14 +174,50 @@ public void selectMapping(String mappingName, SelenideElement containerElement)
...
+                }
+                if(detailElement.exists()) {
+                    detailElement.shouldBe(visible);
+                } else {
+                    containerElement.find(By.cssSelector("div[class=\"DocumentFieldDetailComponent\"")).shouldBe(visible).click();
+                    detailElement.shouldBe(visible);
+                }
+            }
+
             if (detailElement.find(Element.CHILDREN).exists()) {
```
asmigala: I know this is not your code but since we are touching it already, could you flip the condition here to prevent empty `if`?

mastepan:
retest this please

ui-tests/src/test/java/io/syndesis/qe/pages/integrations/editor/add/steps/DataMapper.java
```diff
@@ -174,14 +174,50 @@ public void selectMapping(String mappingName, SelenideElement containerElement)
         List<String> path = Arrays.asList(mappingName.trim().split("\\."));
         final String parent = path.get(0);
 
-        path.forEach(s -> {
-            SelenideElement detailElement = containerElement.find(By.id(s)).shouldBe(visible);
+        boolean previousHadNullId = false;
+
+        for( int i = 0; i < path.size(); i++ ) {
+            String s = path.get(i);
+            SelenideElement detailElement = containerElement.find(By.id(s));
```
asmigala: We try to find the `detailElement` here, but later we check that `s` was either null or empty string and then skip everything else. I believe we could move finding the `detailElement` after the `if` below.

ui-tests/src/test/java/io/syndesis/qe/pages/integrations/editor/add/steps/DataMapper.java
```diff
@@ -174,14 +174,50 @@ public void selectMapping(String mappingName, SelenideElement containerElement)
...
+
+            if (s == null || s.equals("")) {
```
asmigala: Could the `s` actually ever be null? The `String.split` javadoc does not say anything about that, only mentions empty strings.

Also, we currently `find(By.id(s))`, which would immediately throw `IllegalArgumentException` if `s` was null.

syndesisqe:
Build finished. Tests FAILED

mastepan:
retest this please

syndesisqe:
Build finished. Tests PASSED

ui-tests/src/test/java/io/syndesis/qe/pages/integrations/editor/add/steps/DataMapper.java
mastepan: Another good practice is having positive conditions in Ifs. I'd leave it as it is. It's well readable.

mastepan:
@asmigala I've gone through all your comments and fixed the issues. This is ready to merge if you agree.

syndesisqe:
Build finished. Tests FAILED

mcada:
@mastepan I went through test results from latest master run and there is not a single example of test failing due to data mapper empty collection. Unless you show us your test and an example where it fails now and does not with this fix, this PR will not be merged.

avano:
@mcada just a note that since there are no new daily builds for 2 weeks, nightly job is using syndesis from 20.5.2019

mcada:
@avano syndesis version does not really influence this PR as our problem is whether we merge this one and make it almost not readable algorithm or we go with simpler solution which does not work in a specific use-case. The use-case is not present in syndesis for now (Martin did not provide an example) and if it is in FHIR connector, we would have to merge this one and think about refactoring whole mapping step later.

avano:
> test results from latest master run and there is not a single example of test failing due to data mapper empty collection

I was replying to this

mcada:
As I mentioned on the other competing PR to clarify our requirement: what we want to see is a mapping with unnamed collection inside data mapper which is not on position 1. If it is in your FHIR test, please do a draft PR so we can check it out, its as simple as that :)

mcada:
@avano yes I wanted to emphasize that I went through all the tests and did not see any issue where our mapping step in master testsuite which currently opens all the collections caused an error. I was wrong as there is at least one - `Integration - Twitter to Database` but the cause of the fail is different - two similar fields and thus wrong mapping done. This is fixed with https://github.com/syndesisio/syndesis-qe/pull/938

mcada:
Based on discussion with Martin - https://github.com/syndesisio/syndesis-qe/pull/938 fixes his use case so we will go with simpler solution. If we encounter a situation where only mapping algorithm change helps - we will revert https://github.com/syndesisio/syndesis-qe/pull/938 and go with this one.

PR #964, opened 2019-06-10
Dropdowns and DM multiple mappings fix

jsafarik:
Added support for button dropdowns (only in fillBy method) & Added scroll back to plus sign in data mapper

//skip-ci


syndesisqe:
Build finished. Tests PASSED

PR #1068, opened 2019-08-22
[UI][fix] - fixed randomly failing 'Validate all credentials' smoke test

mkralik3:
Sometimes the 'Validate all credentials smoke' test fails in PR job. I noticed in the code that we find all forms in the page, filter the particular form and check when the form is there. Probably the page has not been ready yet when the code was looking for the form.
Does we need to do that way? When we know the particular cssClass name for the form, we can just find it.


syndesisqe:
Build finished. Tests PASSED

PR #1080, opened 2019-09-11
[fix] adapted latest changes in ContinuousDeliveryEnvironment

mkralik3:
//skip-ci


syndesisqe:
Build finished. Tests PASSED

mkralik3:
@avano I remove WIP label because the public api rest tests need to be updated (according to changes in the server) anyway. So this fix fixes only compilation error and not public api rest tests.

PR #1119, opened 2019-09-25
[FUSEQE-6153] Temporary disable upgrade tests

avano:
//skip-ci


syndesisqe:
Build finished. Tests PASSED

tplevko:
You can also already use "@ENTESB-xxxxx" label for the tests to indicate there is related Jira :)

avano:
the tests won't work unless I rework them, so it probably makes no sense to add a jira annotation there

PR #1136, opened 2019-10-01
[FUSEQE-5666] Assure that the poll will happen only once at start

avano:
//test: `@box`


syndesisqe:
Build finished. Tests PASSED

PR #1145, opened 2019-10-01
[Broker] Rename broker so that it won't collide with the one from pro…

avano:
…d build

//skip-ci


syndesisqe:
Build finished. Tests PASSED

PR #1162, opened 2019-10-04
[tw_sf] Fix for second set of credentials

tplevko:
//test: `@integrations-twitter-to-salesforce`

Updated this test, so it will work also with the second set of credentials


syndesisqe:
Build finished. Tests PASSED

PR #1168, opened 2019-10-09
[DB-to-DB] Fix smoke test - remove unnecessary open collection mapping

tplevko:
should run only smokes


syndesisqe:
Build finished. Tests PASSED

PR #1204, opened 2019-10-21
Classes are transformed only when selector reporting is enabled

mmuzikar:
//skip-ci
The testsuite logged that it's generating selector reports even if that wasn't true, also I refactored ClassTransformerHook a bit to make it future proof if we want to add any more transformers.

### DIFF
--- a/ui-tests/src/test/java/io/syndesis/qe/hooks/ClassTransformerHook.java
+++ b/ui-tests/src/test/java/io/syndesis/qe/hooks/ClassTransformerHook.java
@@ -1,5 +1,6 @@
 package io.syndesis.qe.hooks;
 
+import io.syndesis.qe.TestConfiguration;
 import io.syndesis.qe.utils.ExcludeFromSelectorReports;
 
 import cucumber.api.event.EventListener;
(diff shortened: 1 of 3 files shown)


syndesisqe:
Build finished. Tests PASSED

ui-tests/src/test/java/io/syndesis/qe/hooks/ClassTransformerHook.java
```diff
@@ -71,6 +72,13 @@ private void transform() {
             }).installOnByteBuddyAgent();
     }
 
+    private void transform() {
+        ByteBuddyAgent.install();
```
asmigala: Any reason this is not also inside the if?

syndesisqe:
Build finished. Tests PASSED

ui-tests/src/test/java/io/syndesis/qe/hooks/ClassTransformerHook.java
mmuzikar: There was -> it should only be called once, but I changed it to lazy load it for each transformer.

PR #1207, opened 2019-10-22
[Install] Add a check for route condition

avano:
if there is a problem with the route, it ends with:

```
java.lang.AssertionError: Syndesis route failed to provision because of: a route in another namespace holds avano-syndesis.my-minishift.syndesis.io and is older than syndesis
```

//skip-ci


syndesisqe:
Build finished. Tests PASSED

PR #1214, opened 2019-10-24
[ui] Add MongoDB happy path tests

asmigala:
//test: `@mongodb`

### DIFF
--- a/ui-tests/src/test/resources/features/integrations/mongodb.feature
+++ b/ui-tests/src/test/resources/features/integrations/mongodb.feature
@@ -1,0 +1,387 @@
+# @sustainer: asmigala@redhat.com
+
+@ui
+@mongodb
+@database
+@integrations-mongodb
+Feature: MongoDB
+
+  Background: Clean application state
+    Given clean application state
+    And deploy MongoDB 3.6 database
+    And connect to MongoDB "mongodb36"
+    And log into the Syndesis
+    And created connections
+      | MongoDB | mongodb36 | Mongo36 | Mongo description |
+
+
+  @mongodb-insert
+  @mongodb-consume
+  Scenario: MongoDB insert
+    Given create mongodb capped collection "capped_for_insert" with size 1000 and max 3
+    And create mongodb collection "insert_collection"
+    When navigate to the "Home" page
+    And click on the "Create Integration" link to create a new integration.
+    Then check visibility of visual integration editor
+    And check that position of connection to fill is "Start"
+
+    When select the "Mongo36" connection
+    And select "Mongo consumer" integration action
+    And fill in values by element data-testid
+      | collection               | capped_for_insert |
+      | tailtrackincreasingfield | _id               |
+    And click on the "Next" button
+
+    Then check visibility of page "Choose a Finish Connection"
+
+    When select the "Log" connection
+    And fill in values by element data-testid
+      | bodyloggingenabled | true |
+    And click on the "Next" button
+
+    And add integration step on position "0"
+    And select the "Mongo36" connection
+    And select "Mongo producer" integration action
+    And fill in values by element data-testid
+      | collection | insert_collection |
+      | operation  | Insert            |
+    And click on the "Next" button
+
+    And publish integration
+    And set integration name "Mongo to Mongo insert"
+    And publish integration
+
+    Then navigate to the "Integrations" page
+    And wait until integration "Mongo to Mongo insert" gets into "Running" state
+
+    When insert the following documents into mongodb collection "capped_for_insert"
+      | _id | value |
+      | id1 | v1    |
+
+    Then verify that mongodb collection "insert_collection" has 1 document matching
+      | _id | value |
+      | id1 | v1    |
+    And validate that logs of integration "Mongo to Mongo insert" contains string "\"value\" : \"v1\""
+
+
+  @mongodb-update
+  Scenario: MongoDB update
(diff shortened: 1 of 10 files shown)


syndesisqe:
Build finished. Tests PASSED

utilities/src/main/java/io/syndesis/qe/templates/MongoDb36Template.java
```diff
@@ -0,0 +1,132 @@
...
+            .withName(APP_NAME)
+            .addToLabels(LABEL_NAME, APP_NAME)
+            .endMetadata()
+            .editOrNewSpecLike(serviceSpecBuilder.build())
+            .endSpec()
+            .done();
+
+        try {
+            OpenShiftWaitUtils.waitFor(OpenShiftWaitUtils.areExactlyNPodsReady(LABEL_NAME, APP_NAME, 1));
+            Thread.sleep(20 * 1000);
```
avano: why this sleep?

ui-tests/src/test/resources/features/integrations/mongodb.feature
```diff
@@ -0,0 +1,386 @@
...
+    And set integration name "Mongo to Mongo insert"
+    And publish integration
+
+    Then navigate to the "Integrations" page
+    And wait until integration "Mongo to Mongo insert" gets into "Running" state
+
+    When insert the following documents into mongodb collection "capped_for_insert"
+      | _id | value |
+      | id1 | v1    |
+    And sleep for 5 seconds
```
avano: woudn't it make more sense to have some wait in `verify that mongodb collection ...` rather than a random value here?

avano:
@asmigala also something went wrong with the first job execution, it is executing for 15h now and it seems that it can't terminate the integrations:

```
i-mongo-to-mongo-insert-1-xxkp6     0/1     Terminating   0          14h
i-webhook-to-mongo-update-2-qb8v7   0/1     Terminating   0          14h
```

from the server log, these lines are printer over and over again:

```
2019-10-25 07:01:21.728  INFO [-,,,] 1 --- [ning]: pollPods] i.s.s.l.j.c.ActivityTrackingController   : Recovered state: {}
2019-10-25 07:01:21.730  INFO [-,,,] 1 --- [-insert-1-xxkp6] i.s.s.l.j.c.ActivityTrackingController   : Getting controller for pod: i-mongo-to-mongo-insert-1-xxkp6
2019-10-25 07:01:21.734  INFO [-,,,] 1 --- [timestamps=true] i.s.s.l.j.controller.KubernetesSupport   : Failure occurred while processing controller for pod: i-mongo-to-mongo-insert-1-xxkp6, http status: 400, details: {"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"container \"i-mongo-to-mongo-insert\" in pod \"i-mongo-to-mongo-insert-1-xxkp6\" is terminated","reason":"BadRequest","code":400}

```

did you encounter any of that when developing the tests? :thinking:

syndesisqe:
Build finished. Tests FAILED

utilities/src/main/java/io/syndesis/qe/templates/MongoDb36Template.java
asmigala: honestly don't know, most of this is copy-pasta from other templates, will remove

ui-tests/src/test/resources/features/integrations/mongodb.feature
asmigala: Hmm, the problem here is that we put values into one mongo collection, the integration takes that and puts it into the other collection and then we verify. The delay is not in the verification, but in the time it takes the mongo connector to realize there's new data in the tracked collection.

Adding a wait in either the insert above or the verify below would mean we would have to way every time, even though it's only neccessary in this case.

Maybe I can add a comment to that effect here?

asmigala:
> did you encounter any of that when developing the tests? 

I did encounter some weird behaviour (mostly 500 response from reset db over and over again). I _think_ this is because of the port forward, since that's the change I did between the first and second test run. I'll do the changes you requested  and we'll see how the tests go then.

syndesisqe:
Build finished. Tests FAILED

ui-tests/src/test/resources/features/integrations/mongodb.feature
asmigala: So the transfer from one collection to other should be almost instant, because the connector is subscribed to the changes, it's just that this code is even faster. 

I'll add a wait, but I'll have to spend some more time on this, since using the `waitFor` would lose me the nice assertion message I have when the doc is not found

avano: yeah, the keywords `should be almost instant` are not 100% true on jenkins :grin:

asmigala: The `almost instant` part happens between two pods in the same namespace in openshift, so in this instance it's the code running on jenkins which _could_ be too fast. At least it was _sometimes_ when running locally, I guess it would work ok on jenkins.

avano: > I guess it would work ok on jenkins

so lets leave it this way and we will see

syndesisqe:
Build finished. Tests FAILED

syndesisqe:
Build finished. Tests PASSED

asmigala:
@avano should be good to go

PR #1221, opened 2019-10-29
[TestSupport] Revert cleaning RCs

avano:
//skip-ci

it's *possible* that cleaning up the RCs causes the pods to stay in terminating state *sometimes*


syndesisqe:
Build finished. Tests PASSED

PR #1234, opened 2019-11-12
[SQS] Retry purge queue after delay if there is some already in progress

avano:
//skip-ci

### DIFF
--- a/utilities/src/main/java/io/syndesis/qe/utils/SQSUtils.java
+++ b/utilities/src/main/java/io/syndesis/qe/utils/SQSUtils.java
@@ -20,6 +20,7 @@ import software.amazon.awssdk.auth.credentials.AwsBasicCredentials;
 import software.amazon.awssdk.regions.Region;
 import software.amazon.awssdk.services.sqs.SqsClient;
 import software.amazon.awssdk.services.sqs.model.Message;
+import software.amazon.awssdk.services.sqs.model.PurgeQueueInProgressException;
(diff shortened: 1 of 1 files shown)


syndesisqe:
Build finished. Tests PASSED

utilities/src/main/java/io/syndesis/qe/utils/SQSUtils.java
```diff
@@ -149,6 +150,11 @@ public void purge(String... queueNames) {
                 client.purgeQueue(b -> b.queueUrl(getQueueUrl(queueName)).build());
             } catch (QueueDoesNotExistException e) {
                 // ignore
+            } catch (PurgeQueueInProgressException ex) {
+                // If for some reason some other purge queue is in progress, wait and retry
+                log.debug("Purging " + queueName + " threw PurgeQueueInProgressException, waiting and retrying");
+                TestUtils.sleepIgnoreInterrupt(90000L);
+                client.purgeQueue(b -> b.queueUrl(getQueueUrl(queueName)).build());
```
asmigala: Would it make sense to keep retrying in a loop until successful, or is this such an edge case that it's not necessary?

avano: to be honest, I really don't have any idea why it occasionally fails with that exception, as there should not be any purge in progress, so I would say it is really an edge case

PR #1253, opened 2019-11-28
[soak-test] fix delay problem

mmelko:
//skip-ci
add reset services before scenario start


syndesisqe:
Build finished. Tests PASSED

PR #1289, opened 2020-01-17
[ENTESB-12493] Add ENTESB-12493 reproducer

avano:
//skip-ci

skip ci as the whole feature expects stored procedures to be working, but currently they are not, so it would fail anyway


syndesisqe:
Build finished. Tests PASSED

PR #1291, opened 2020-01-17
[servicenow] Fix get incident from service now

tplevko:
//test: `@integrations-servicenow-1`


syndesisqe:
Build finished. Tests PASSED

PR #1299, opened 2020-01-27
[FUSEQE-7714] fixed login on OCP4.3

mmuzikar:


syndesisqe:
Build finished. Tests PASSED

avano:
maybe it is a good time to start to check what is on the page and not check the url, wdyt?

mmuzikar:
@avano that could be easily done, but I am not sure about the other logging methods DOM structure so it doesn't fill in other forms.

avano:
my guess would be that at least 2/4 are not used anymore :dancer: maybe @mcada would know what login methods we need, as far as I know we test our clusters and then the staging, but I don't know how the login looks like there

mcada:
I am more concerned with login not working somewhere then having code which will not be executed. Minishift, CRC, OCP3.x, OCP4.x and staging cluster are the only environments I recall we use. Some consolidation should be done here in the future.

mmuzikar:
So @mcada @avano do I understand that this is fine to merge as temporary fix and it will get properly fixed when there's more time?

PR #1307, opened 2020-01-31
[Dropbox] Catch exception when deleting file that is not present

avano:
//skip-ci


syndesisqe:
Build finished. Tests PASSED

PR #1313, opened 2020-02-05
[CRD] Another fix for existing CRD but in different version

avano:
//skip-ci

### DIFF
--- a/utilities/src/main/java/io/syndesis/qe/resource/impl/Syndesis.java
+++ b/utilities/src/main/java/io/syndesis/qe/resource/impl/Syndesis.java
@@ -286,7 +286,28 @@ public class Syndesis implements Resource {
                     .endVersion()
                     // Edit the other version (Let's hope that there will be max 2 versions concurrently)
                     .editMatchingVersion(v -> !v.getName().equals(newCrd.getSpec().getVersion()))
-                    .withServed(false)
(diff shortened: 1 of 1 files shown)


syndesisqe:
Build finished. Tests PASSED

utilities/src/main/java/io/syndesis/qe/resource/impl/Syndesis.java
```diff
@@ -286,7 +286,28 @@ public void deployCrd() {
...
+                // We need to make "current" CRD version "served" and with "storage"
+                OpenShiftUtils.getInstance().customResourceDefinitions().withName(existingCrd.getMetadata().getName())
+                    .edit()
+                    .editSpec()
+                    // Edit the version we want to deploy now
+                    .editMatchingVersion(v -> v.getName().equals(newCrd.getSpec().getVersion()))
+                    .withServed(true)
+                    .withStorage(true)
+                    .endVersion()
+                    // Edit the other version (Let's hope that there will be max 2 versions concurrently)
```
asmigala: > Let's hope that there will be max 2 versions concurrently
:D :D :D

avano: :crossed_fingers:

PR #1322, opened 2020-02-11
[fix] - remove security selection

mkralik3:
There is only one connector security type and the item cannot be modified

//test: `@integrations-amq-to-rest or @integrations-webhook-api-connector`


syndesisqe:
Build finished. Tests FAILED

PR #1337, opened 2020-02-19
[Deploy] Add lastUsedBy annotation to the project

avano:
//skip-ci

### DIFF
--- a/utilities/src/main/java/io/syndesis/qe/TestSuiteParent.java
+++ b/utilities/src/main/java/io/syndesis/qe/TestSuiteParent.java
@@ -8,6 +8,8 @@ import io.syndesis.qe.utils.TestUtils;
 import org.junit.AfterClass;
 import org.junit.BeforeClass;
 
+import java.util.Map;
+
 import lombok.extern.slf4j.Slf4j;
 
 @Slf4j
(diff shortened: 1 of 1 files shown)


syndesisqe:
Build finished. Tests PASSED

utilities/src/main/java/io/syndesis/qe/TestSuiteParent.java
```diff
@@ -41,6 +43,17 @@ public static void beforeTests() {
             TestUtils.sleepIgnoreInterrupt(10 * 1000L);
         }
 
+        // You can't create project with annotations/labels - https://github.com/openshift/origin/issues/3819
+        // So add them after the project is created
+        // @formatter:off
+        Map<String, String> annotations = TestUtils.map("syndesis-qe/lastUsedBy", System.getProperty("user.name"));
+        OpenShiftUtils.getInstance().namespaces().withName(TestConfiguration.openShiftNamespace()).edit()
+            .editMetadata()
+                .addToAnnotations(annotations)
```
asmigala: I'm not 100 % sure about the difference between labels and annotations but I found that labels are for things seen by users and annotations are for internal purposes. Would it make more sense to add this as a label?

avano: that's exactly why I put it in the annotations :grin: , but I can add it to labels, no problem

syndesisqe:
Build finished. Tests PASSED

PR #1343, opened 2020-02-24
Fix api connector tests

mastepan:
//test: `@DB-custom-api-connector-DB`


syndesisqe:
Build finished. Tests PASSED

mkralik3:
The issue with HTTP Basic (when it is already selected and the element is disabled) was already fixed here: ec6df4a#diff-6dba28b49f11b0d1e47181677e5db8a7R88 I have tried it now on master and this test works correctly.
In my POV, I would leave that step in the feature file in case that the security type will have to be selected manually in some future Syndesis version. WDYT? If you agree, please close this issue.

mastepan:
I'm closing it as it works on master

PR #1344, opened 2020-02-24
Fix db to db tests

mastepan:
//test: `@db-insert-multiple-rows`


mkralik3:
The issue with HTTP Basic (when it missing) was already fixed here: https://github.com/syndesisio/syndesis-qe/commit/ec6df4a3125cad183a208cb84ad97b68b798d5e5#diff-6dba28b49f11b0d1e47181677e5db8a7R88  I have tried it now on master and this test works correctly. 
In my POV, I would leave that step in the feature file in case that the security type will have to be selected manually in some future Syndesis version.

syndesisqe:
Build finished. Tests PASSED

mastepan:
Closing as it works on master

PR #1374, opened 2020-03-17
Log improvement

mkralik3:
//skip-ci
`waitForSyndesis` is already in the `deploySyndesis` method. I moved the log to that method in order to it is shown in the correct phase.


syndesisqe:
Build finished. Tests PASSED

PR #1466, opened 2020-05-05
[ENTESB-11455] Remove verification for non-existant response body

mmuzikar:
//test: `@ENTESB-11455`
The response body is null, so checking for it to be empty causes errors, the important detail is the response code.


syndesisqe:
Build finished. Tests PASSED

PR #1471, opened 2020-05-13
[FUSEQE-8862] DataMapper migration

mkralik3:
//skip-ci

DataMapper migration.
PR contains also a commit which deletes unused steps since the data buckets are expanded by default and the 'create data mapper mappings' step explicitly expands all collections too.
It does not need to do it manually anymore.


syndesisqe:
Build finished. Tests PASSED

tplevko:
Went through the changes, seems to be all right. We can merge once https://github.com/syndesisio/syndesis/pull/8456 get's merged.

PR #1474, opened 2020-05-14
[FIX] Fix sveres testsuite

Stefan365:
actually backport to 1.9.x


mkralik3:
@Stefan365 Is the commit message correct?

PR #1478, opened 2020-05-18
[ResetDB] Reset DV when it is present

avano:


syndesisqe:
Build finished. Tests PASSED

PR #1481, opened 2020-05-18
[FUSEQE-8862] Support for fields with the same name in DataMapper

mkralik3:
* Added support for same field name in the different data buckets
* Migrated conditional flow
* Added blocker annotation

//test: `@dynamodb-query or @integrations-conditional-flows`


syndesisqe:
Build finished. Tests PASSED

PR #1492, opened 2020-05-25
[FUSEQE-9074] create automated tests for ENTESB-12882

AnetaCadova:
//test: `@check-metering-labels`

### DIFF
--- a/rest-tests/src/test/java/io/syndesis/qe/rest/tests/steps/OperatorMeteringLabels.java
+++ b/rest-tests/src/test/java/io/syndesis/qe/rest/tests/steps/OperatorMeteringLabels.java
@@ -1,0 +1,36 @@
+package io.syndesis.qe.rest.tests.steps;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import io.syndesis.qe.utils.OpenShiftUtils;
+
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+import cucumber.api.java.en.Then;
+import io.fabric8.kubernetes.api.model.Pod;
+
+public class OperatorMeteringLabels {
+    /**
+     * Checks whether pods contains com.redhat metering labels.
+     * It is a feature for Fuse Online product, therefore check runs only in case of the productized build.
+     */
+    @Then("verify new RedHat metering labels")
+    public void checkRedhatLabels() {
+        List<Pod> pods = OpenShiftUtils.getInstance().pods().withLabel("syndesis.io/component").list().getItems().stream()
+            .filter(p -> !"integration".equals(p.getMetadata().getLabels().get("syndesis.io/component")))
+            .collect(Collectors.toList());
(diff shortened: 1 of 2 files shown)


syndesisqe:
Build finished. Tests PASSED

rest-tests/src/test/java/io/syndesis/qe/rest/tests/steps/OperatorMeteringLabels.java
```diff
@@ -0,0 +1,32 @@
...
+import io.fabric8.kubernetes.api.model.Pod;
+
+public class OperatorMeteringLabels {
+    /**
+     * Checks whether pods contains com.redhat metering labels.
+     * It is a feature for Fuse Online product, therefore check runs only in case of the productized build.
+     */
+    @Then("verify new RedHat metering labels")
+    public void checkRedhatLabels() {
+        if (TestUtils.getSyndesisVersion().contains("redhat")) {
```
mkralik3, line 21: You can use @prod test annotation (https://github.com/syndesisio/syndesis-qe/blob/1.10.x/rest-tests/src/test/java/io/syndesis/qe/RestTestHooks.java#L38) instead of if statement in order to run the test only with the prod builds.

syndesisqe:
Following scenarios failed:
```
Integration - DB to DB | Read and update operations 
```

mkralik3:
@AnetaCadova I have tried to run this test locally against product version and it failed on
```
Then verify new RedHat metering labels # OperatorMeteringLabels.checkRedhatLabels()
      java.lang.AssertionError: 
Expecting:
 <{"openshift.io/build.name"="todo-1"}>
to contain key:
 <"com.redhat.product-name">
	at io.syndesis.qe.rest.tests.steps.OperatorMeteringLabels.checkRedhatLabels(OperatorMeteringLabels.java:23)
	at ✽.verify new RedHat metering labels(classpath:features/check-metering-labels.feature:12)

```

The pod `todo-1-build` doesn't have any tested labels. Should it have them? If yes, it is an issue, if not, the test needs an update. 

Also, I am not sure if calling `OpenShiftUtils.getInstance().getPods()` method is safe. What if some pods from the previous tests will be hanging in the namespace that would not RedHat metering labels? (which would cause test failure)

I don't know the background of ENTESB-12882. What pods should be labeled by those labels?
If only Syndesis component pods, you can use something similar as in this [method](https://github.com/syndesisio/syndesis-qe/blob/1.10.x/utilities/src/main/java/io/syndesis/qe/Component.java#L71) (but that method excludes all build and deploy pods, also integration pods. So before you use it, make sure that you don't need them for the test. Otherwise, create a new one more suitable for your test. Also, that method doesn't return jaeger-operator pod so you need to get them differently (e.g. by name :D ))

In case the integration also contains mettering labels, I suggest to extend test about create and publish simple integration and test integration pod too.

syndesisqe:
Build finished. Tests PASSED

PR #1510, opened 2020-06-03
[FUSEQE-9292] Update activity error tests

mkralik3:
//test: `@activity-error or @activity-error-data-mapper-before or @ENTESB-13988`


mkralik3:
@mmuzikar 

Tests for codes 200 ,201, 202 and 400 should fail due to ENTESB-13988

syndesisqe:
Following scenarios failed:
```
Webhook extension | Check whether webhook returns 400 error code 
```

PR #1512, opened 2020-06-09
FUSEQE-9312 Fix 10-integrations-complex test and replace odata with servicenow

mkralik3:
//skip-ci

PR #1514, opened 2020-06-09
Backport master

mkralik3:
Backports for:
https://github.com/syndesisio/syndesis-qe/pull/1512
https://github.com/syndesisio/syndesis-qe/pull/1513


syndesisqe:
Following scenarios failed:
```
Integration info | Check error for integration with the same name 
Integration info | Check error for integration which has same name after update 
```

PR #1517, opened 2020-06-10
[PortForward] Handle not working port forward and select only running…

avano:
… pods

//skip-ci


syndesisqe:
Build finished. Tests PASSED

PR #1542, opened 2020-06-23
[FUSEQE-9414] Fix datamapper instability on Chrome

mkralik3:
* when the constant/property modalDialog is closed, the tooltip for adding, which was opened before, remains open. It may cause instability on chrome. Added click on root element (the tooltip is disappeared)
* clarified log message for Jaeger SA


syndesisqe:
Build finished. Tests PASSED

PR #1550, opened 2020-07-07
[hotfix] datamapper

mkralik3:
//skip-ci

DM select the field with contains, it caused the failure that when the user wants to map `text` field and DM contains e.g. following fields
```
text
field2text
text22
```
Also, there cannot be used equal method, because fields in the property data bucket contains also default value in the name e.g.
```
text = default
```


syndesisqe:
Build finished. Tests PASSED

PR #1555, opened 2020-07-08
Fix FHIR tests according to changes in DataMapper

mkralik3:
* Fixed collection name collision in DataMapper.

//skip-ci

Fix FHIR tests according to changes in DataMapper ( https://github.com/syndesisio/syndesis-qe/pull/1550 )
Fixed collection name collision in DataMapper.
E.g. when `id` collection should be selected, more names met the condition, e.g.  `id` and `identifier` , which caused the failure.
Now, the collection name is directly specified in DataTestId


syndesisqe:
Build finished. Tests PASSED

PR #1601, opened 2020-09-02
Fix wait for Jaeger operator in the second namespace

mkralik3:

### DIFF
--- a/utilities/src/main/java/io/syndesis/qe/resource/impl/Syndesis.java
+++ b/utilities/src/main/java/io/syndesis/qe/resource/impl/Syndesis.java
@@ -163,9 +163,9 @@ public class Syndesis implements Resource {
     public void jaegerWorkarounds() {
         new Thread(() -> {
             try {
-                OpenShiftWaitUtils.waitUntilPodAppears("jaeger-operator");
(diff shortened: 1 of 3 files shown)


utilities/src/main/java/io/syndesis/qe/resource/impl/Syndesis.java
```diff
@@ -163,7 +163,7 @@ public void createPullSecret() {
     public void jaegerWorkarounds() {
         new Thread(() -> {
             try {
-                OpenShiftWaitUtils.waitUntilPodAppears("jaeger-operator");
+                OpenShiftWaitUtils.waitUntilPodAppearsOrAnotherPodPresent("jaeger-operator", "syndesis-jaeger");
```
avano, line 166: maybe it would be better to change `waitUntilPodAppears(String podName)` into `waitUntilAnyPodAppears(String... podNames)`, wdyt?

mkralik3: Yes, it would be probably better :D

syndesisqe:
Build finished. Tests PASSED

avano (approved):

syndesisqe:
Build finished. Tests PASSED

PR #1653, opened 2020-11-16
[FUSEQE-10617] Operatorhub automation

mmuzikar:

### DIFF
--- a/utilities/src/main/java/io/syndesis/qe/resource/impl/Syndesis.java
+++ b/utilities/src/main/java/io/syndesis/qe/resource/impl/Syndesis.java
@@ -4,9 +4,15 @@ import static org.assertj.core.api.Assertions.fail;
 
 import io.syndesis.qe.TestConfiguration;
 import io.syndesis.qe.addon.Addon;
+import io.syndesis.qe.common.CommonSteps;
 import io.syndesis.qe.component.Component;
 import io.syndesis.qe.component.ComponentUtils;
 import io.syndesis.qe.image.Image;
+import io.syndesis.qe.marketplace.manifests.Bundle;
+import io.syndesis.qe.marketplace.manifests.Index;
+import io.syndesis.qe.marketplace.manifests.Opm;
+import io.syndesis.qe.marketplace.openshift.OpenShiftService;
+import io.syndesis.qe.marketplace.quay.QuayUser;
 import io.syndesis.qe.resource.Resource;
 import io.syndesis.qe.resource.ResourceFactory;
 import io.syndesis.qe.test.InfraFail;
@@ -87,11 +93,22 @@ public class Syndesis implements Resource {
         log.info("Deploying Syndesis");
         log.info("  Cluster:   " + TestConfiguration.openShiftUrl());
         log.info("  Namespace: " + TestConfiguration.openShiftNamespace());
+
+        if (TestConfiguration.getIndexImage() != null || TestConfiguration.getBundleImage() != null) {
+            deployViaBundle();
+            return;
+        }
+
         createPullSecret();
         pullOperatorImage();
         installCluster();
-        grantPermissions();
         deployOperator();
+        deployCrAndRoutes();
+    }
+
+    public void deployCrAndRoutes() {
+        log.info("Deploying Syndesis CR");
+        grantPermissions();
         deploySyndesisViaOperator();
         changeRuntime(TestConfiguration.syndesisRuntime());
         checkRoute();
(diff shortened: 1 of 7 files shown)


syndesisqe:
Build finished. Tests PASSED

mmuzikar:
@avano I am wondering whether we should also make something similar to -Pdeploy or add this logic to -Pdeploy. What do you think?

avano:
> @avano I am wondering whether we should also make something similar to -Pdeploy or add this logic to -Pdeploy. What do you think?

:thinking: what would you want to deploy?

mmuzikar:
Syndesis subscription on Operatorhub

avano:
but that won't be used outside of the operatorhub tests, will it?

mmuzikar:
yes that'd be just a convenience feature

avano:
maybe it would be better to move the code for deploying from operatorhub into `Syndesis` class and then decide based on the properties set what deploy method to use?

like, if you have `syndesis.config.bundle.image` `syndesis.config.index.image` set -> use operator hub, otherwise use the standard method. wdyt?

syndesisqe:
Build finished. Tests PASSED

syndesisqe:
Build finished. Tests PASSED

syndesisqe:
Build finished. Tests PASSED

avano:
retest this please

avano:
retest this please

avano:
retest this please

avano:
retest this please

avano:
retest this please

syndesisqe:
Following scenarios failed:
```
Integration - Database | Smoke - Periodic invocation to Insert 
```

mkralik3:
Is step `Given deploy Syndesis via bundle` defined somewhere?  I am getting `The step "deploy Syndesis via bundle" is undefined.`

mmuzikar:
@mkralik3 We decided to remove the scenario and rather use the deploy profile to setup operatorhub. I updated the README as well to make this clear.

syndesisqe:
Following scenarios failed:
```
Integration - DB to DB | Read and update operations 
```

avano:
retest this please

syndesisqe:
Build finished. Tests PASSED

utilities/src/main/java/io/syndesis/qe/resource/impl/Syndesis.java
```diff
@@ -705,4 +776,113 @@ public void linkServiceAccountWithSyndesisPullSecret(String serviceAccountName)
...
+            operatorDeployment.getJSONObject("spec").getJSONObject("template").getJSONObject("spec").getJSONArray("containers").getJSONObject(0)
+                .getJSONArray("env");
+        envVars.put(TestUtils.map("name", "TEST_SUPPORT", "value", "true"));
+        try {
+            OpenShiftUtils.getInstance().customResource(getCSVContext())
+                .edit(TestConfiguration.openShiftNamespace(), "fuse-online-operator.v7.8.0", json.toMap());
+        } catch (IOException e) {
+            log.error("Couldn't edit Syndesis CSV", e);
+        }
+    }
```
avano, line 887: there is a code for that here: https://github.com/syndesisio/syndesis-qe/blob/master/utilities/src/main/java/io/syndesis/qe/TestSuiteParent.java#L37-L60

please merge it into one method that will do the following:
- TestSuiteParent will call that method
- in that method, check what type of deployment it is - either "standard" or operator hub
- for operator hub, check if the test support is already enabled, if no, enable it
- for standard deployment, check `if TestConfiguration.enableTestSupport()` and if yes, do that part of the code

utilities/src/main/java/io/syndesis/qe/resource/impl/Syndesis.java
```diff
@@ -444,6 +480,41 @@ public void deployOperator() {
...
+            TestUtils.withRetry(() -> {
+                try {
+                    OpenShiftUtils.getInstance().updateDeploymentConfigEnvVars(operatorResourcesName, finalImagesEnvVars);
+                    return true;
+                } catch (KubernetesClientException kce) {
+                    log.debug("Caught KubernetesClientException: " + kce);
+                    return false;
+                }
+            }, 3, 30000L, "Unable to update operator deployment config after 3 tries");
+        }
```
avano, line 517: :thinking: this part was changed in `1.11.x` so please adapt the changes for that

rest-tests/src/test/resources/features/operatorhub.feature
```diff
@@ -0,0 +1,31 @@
...
+
+    And create finish DB invoke sql action step with query "INSERT INTO TODO (task, completed) VALUES (:#task, 3)"
+    Then create integration with name: "DB to DB smoke rest test"
+    Then wait for integration with name: "DB to DB smoke rest test" to become active
+
+    And sleep for jenkins delay or 15 seconds
+
+    Then validate that number of all todos with task "Josef_first" is greater than 0
+    Then validate that number of all todos with task "Josef_second" is 0
+
```
avano, line 31: I don't know how, but there is a missing new line at the end :grin:

avano (changes requested):

utilities/src/main/java/io/syndesis/qe/resource/impl/Syndesis.java
mmuzikar: how does one merge? :smile: 
I think enabling test support needs to happen at different times for both ways. I need to edit the CSV after creating a subscription to ensure CSV is present on the cluster. And this code looks like it needs to run even prior to the operator existing.

utilities/src/main/java/io/syndesis/qe/resource/impl/Syndesis.java
mmuzikar: I guess we could also remove this code and the feature file as the old operatorhub installation won't be used anymore, wdyt?

utilities/src/main/java/io/syndesis/qe/resource/impl/Syndesis.java
avano: you're right, my bad :grin:

utilities/src/main/java/io/syndesis/qe/resource/impl/Syndesis.java
avano: makes sense

syndesisqe:
Following scenarios failed:
```
Integration - DB to DB | Read and update operations 
```

avano (approved):

avano:
retest this please

syndesisqe:
Build finished. Tests PASSED

PR #1691, opened 2021-01-25
Reproducer for ENTESB-15292 and ENTESB-14237

mkralik3:
//test: `@ENTESB-15292`

@fhir-2-2-patch is failing till PR for ENTESB-14237 will be closed

### DIFF
--- a/ui-tests/src/test/resources/features/customizations/connectors/api-client-connector-CRUD.feature
+++ b/ui-tests/src/test/resources/features/customizations/connectors/api-client-connector-CRUD.feature
@@ -47,6 +47,7 @@ Feature: Customization - API Connector CRUD
     Then check visibility of the new connector "Swagger Petstore"
 
   @edit-api-connector
+  @ENTESB-15292
   Scenario: Edit parameters
     When click on the "Customizations" link
     And navigate to the "API Client Connectors" page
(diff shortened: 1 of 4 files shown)


syndesisqe:
Build finished. Tests PASSED

ui-tests/src/test/resources/features/customizations/connectors/api-client-connector-CRUD.feature
```diff
@@ -80,6 +81,14 @@ Feature: Customization - API Connector CRUD
       | api-connector-detail-config-host        | http://petstore.swagger.io-1 |
```
mmuzikar, line 81: Can you check if there's a reason for having steps `validate connector detail values by` and `validate values by element`.
Maybe we can ditch the connector one and use the universal one instead?

mmuzikar (changes requested):

ui-tests/src/test/resources/features/customizations/connectors/api-client-connector-CRUD.feature
mkralik3: Semantically, they are different pages.

The first one is for detail page in API Client Connectors while the second one is for "middle" step in connection creation process (where every connector can define different elements). Even more, the second one has desired text in the atribute `value` while the first one uses text outside of element definition.

However, since the step `validate connector detail values by "data-testid"` is used only here, we can remove it (whole `DetailSteps` class for API Client Connectors) and make some universal step. But I would leave it at it is in case we would like to add there some more steps in case to create some reproducers for API Client Connectors detail page.

mmuzikar: Let's not remove the step then, you are right that the API client connector detail page is subject to change (with authentication and hostnames)

mmuzikar (approved):

PR #1737, opened 2021-04-14
[FUSEQE-12351] - add test for camel header properties

mkralik3:
//skip-ci

Since the DM can have multiple fields with the same name (but different scope) in the Properties data bucket, `ignoreFieldWithSameNameAlreadyMapped` was added to the DM because test suite doesn't support ( fails ) when there are more fields with the same name on the same level in the databucket. Now, when `ignoreFieldWithSameNameAlreadyMapped` is set to true, all fields with the same name, which are already used in DM, are excluded from searching so only one field is found to mapping.
(https://github.com/mkralik3/syndesis-qe/blob/8f0a03da646114ac0d7f89fd85bdbc5a99265a75/ui-common/src/main/java/io/syndesis/qe/pages/integrations/editor/add/steps/DataMapper.java#L289)
Also, there is a limitation that the property has to be mapped before the new property with the same name is created at the same level (otherwise more than one element will be found).


syndesisqe:
Build finished. Tests PASSED

avano (approved):

PR #1743, opened 2021-04-21
[FUSEQE-12433] - improve Google login

mkralik3:
//skip-ci

For some reason, on Firefox in Jenkins node, the Google multi-account table (https://accounts.google.com/accountchooser) is shown during the login even though only the one user was logged before. Due to that, the multi-account table was fixed in the testsuite. Also now, the test suite support more Google accounts during the test run so it is able to choose the correct one from the multi-account table or login a new one through that table if the account is not there.

### DIFF
--- a/ui-common/src/main/java/io/syndesis/qe/steps/CommonSteps.java
+++ b/ui-common/src/main/java/io/syndesis/qe/steps/CommonSteps.java
@@ -52,6 +52,7 @@ import org.openqa.selenium.Dimension;
 import org.openqa.selenium.JavascriptExecutor;
 import org.openqa.selenium.WebDriver;
 import org.openqa.selenium.WebDriverException;
+import org.openqa.selenium.support.ui.WebDriverWait;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.context.annotation.Lazy;
 
@@ -61,6 +62,7 @@ import com.codeborne.selenide.ElementsCollection;
 import com.codeborne.selenide.Selenide;
 import com.codeborne.selenide.SelenideElement;
 import com.codeborne.selenide.WebDriverRunner;
+import com.codeborne.selenide.ex.ElementNotFound;
 
 import java.io.File;
 import java.io.IOException;
@@ -820,13 +822,21 @@ public class CommonSteps {
     private void waitForWindowToShowUrl() {
         TestUtils.waitFor(() -> {
             try {
-                WebDriverRunner.getWebDriver().getCurrentUrl();
+                String currentUrl = WebDriverRunner.getWebDriver().getCurrentUrl();
+                if ("about:blank".equals(currentUrl)) {
+                    return false;
+                }
             } catch (WebDriverException e) {
                 return false;
             }
 
             return true;
         }, 1, 20, "Window has not returned valid url");
+    }
+
+    private void waitTillPageIsLoaded() {
+        new WebDriverWait(WebDriverRunner.getWebDriver(), 300).until(
+            webDriver -> ((JavascriptExecutor) webDriver).executeScript("return document.readyState").equals("complete"));
     }
 
     @ExcludeFromSelectorReports
@@ -861,6 +871,7 @@ public class CommonSteps {
 
         // need to wait a bit until window is able to return url
         waitForWindowToShowUrl();
+        waitTillPageIsLoaded();
 
         if ("firefox".equalsIgnoreCase(TestConfiguration.syndesisBrowser())) {
             WebDriverRunner.getWebDriver().manage().window().setSize(new Dimension(1920, 1080));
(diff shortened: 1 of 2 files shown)


syndesisqe:
Build finished. Tests PASSED

ui-common/src/main/java/io/syndesis/qe/steps/CommonSteps.java
```diff
@@ -933,7 +944,23 @@ private void loginToGoogleIfNeeded(String s) {
         }
 
         waitForCallbackRedirect("google");
-        fillAndValidateGoogleAccount(s);
+
+        Account account = AccountsDirectory.getInstance().get(s);
+        // test if multi accounts table is shown ( https://accounts.google.com/accountchooser )
+        SelenideElement multiAccountsTable = $(By.className("OVnw0d"));
```
mmuzikar: This might break if there's some change on the website. (I suspect that the classnames get randomized with every build). I'm thinking if it might be more future proof to do something like 
```
if ($(By.id("initialView").text().contains("Choose an account")))...
```

ui-common/src/main/java/io/syndesis/qe/steps/CommonSteps.java
```diff
@@ -933,7 +944,23 @@ private void loginToGoogleIfNeeded(String s) {
...
+        if (multiAccountsTable.exists()) {
+            log.info("More Google accounts are available. Choosing the correct one:" + account.getProperty("email"));
+            try {
+                multiAccountsTable.find(ByUtils.customAttribute("data-email", account.getProperty("email"))).should(Condition.exist).click();
+            } catch (ElementNotFound ex) {
+                log.warn("That account is not in the Google multi-accounts table. It needs to fill in the credentials.");
+                $(By.className("UXurCe")).click(); // use another account button
```
mmuzikar: This one might be more future proof with something like $(By.text("Choose an account"))

mmuzikar (commented):
What do you think about using text for the selectors here? I know we tend to avoid it but here it might be the most consistent selector.

ui-common/src/main/java/io/syndesis/qe/steps/CommonSteps.java
mkralik3: I agree that class might not be reliable. There is also `jsname` attribute but I think that can have the same problem.
Problem with text can be that it is changing according to location (or something else). If we can be sure that only English variant will be shown in Jenkins, we can do it that way.
I will look at it more.

mmuzikar: If we want to make sure it is in english the lang selector looks quite reliable `$(By.id("lang-chooser").$(By.data("data-value==en")).click()`

mkralik3: oh, I didn't notice the language chooser :facepalm: . Ok, I am going to update the test suite according to it.

ui-common/src/main/java/io/syndesis/qe/steps/CommonSteps.java
mkralik3: The question is how to implement By.text method since parents contains that text too (first 4 parent element has only that text too so "exact match" still returns 4 elements). Hopefully the root parent with the same text is clickable :D

syndesisqe:
Build finished. Tests PASSED

syndesisqe:
Build finished. Tests PASSED

ui-common/src/main/java/io/syndesis/qe/steps/CommonSteps.java
```diff
@@ -933,7 +944,32 @@ private void loginToGoogleIfNeeded(String s) {
         }
 
         waitForCallbackRedirect("google");
-        fillAndValidateGoogleAccount(s);
+
+        Account account = AccountsDirectory.getInstance().get(s);
+
+        // change language if needs
+        if (!WebDriverRunner.getWebDriver().getCurrentUrl().contains("&hl=en&")) {
```
mkralik3, line 951: When the correct language is already set, the element is not clickable. So we need to check before whether the incorrect language is set.

ui-common/src/main/java/io/syndesis/qe/utils/ByUtils.java
```diff
@@ -34,7 +34,14 @@ public static By rowid(String rowId) {
     }
 
     public static By partialLinkText(String partialText) {
-        return new ByPartialText(partialText);
+        return new By() {
+            @Override
+            public List<WebElement> findElements(SearchContext searchContext) {
+                return searchContext.findElements(By.cssSelector("a,button")).stream()
```
mmuzikar: button can be also <input type="button"> and <input type="submit"> if we want to have all link-ish elements covered

mmuzikar (approved):

syndesisqe:
Build finished. Tests PASSED

PR #1745, opened 2021-04-23
Fix get all tags from quay

mkralik3:
//skip-ci
Quay API for get all tags doesn't have next page url in the response, only the info whether the next page exist. (`has_additional`)


syndesisqe:
Build finished. Tests PASSED

avano (approved):

PR #1762, opened 2021-06-22
Workaround for Jaeger installed by OLM

mkralik3:
On the latest OCP 4.7 instances, we are facing an problem with cleaning project method. When the method OpenShiftUtils.getInstance().clean()`` is called, all resources are deleted but after some time, the OperatorHub create jaeger subscriptions again however the installation stuck in missing deployment. When the Syndesis is being installed right after that (e.g. in the operator tests), it doesn't install Jaeger again (because the Jaeger operator is being installed (actually it is stuck :D )). 

Maybe it has been already somehow fixed in XTF method, but we cannot update XTF until the kubernetes client will be updated in Syndesis ( I need to create a feature request for that)

So, here is this ugly fix


syndesisqe:
Build finished. Tests PASSED

avano (approved):

